PR #2, opened 2022-06-21
Define Product

### LINKED ISSUE #5: Combine and simplify item decorators.
Currently, items are declared like:

```python
import attrs
from zyte_common_items.util import export

@export
@attrs.define(slots=True, kw_only=True)
class SomeItem:
    ...
```

It would be nice if we can combine the two decorators into a single one to have a more concise code.

Gallaecio:
Beyond the actual goal of this pull request: Resolves #5.


zyte_common_items/base.py
```diff
@@ -118,3 +121,52 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+class AdditionalProperty:
+    name: str
+    value: str
+
+
+@export
+@attrs.define(slots=True, kw_only=True)
+class AggregateRating:
+    best_rating: Optional[float] = None
+    rating_value: Optional[float] = None
```
Gallaecio, line 137: In general, for `floats`, do we use `float`, or `Decimal`? `Decimal` would preserve the API output level of detail, although for most fields I am not sure this is important enough.

zyte_common_items/base.py
```diff
@@ -118,3 +121,52 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+@attrs.define(slots=True)
+class Brand:
+    name: str
+
+
+@export
+@attrs.define(slots=True)
+class Breadcrumb:
+    name: str
+    link: str
```
Gallaecio, line 151: `RequestUrl`?

zyte_common_items/items.py
```diff
@@ -0,0 +1,58 @@
...
+    color: Optional[str] = None
+    currency: Optional[str] = None
+    currency_raw: Optional[str] = None
+    gtin: Optional[List[Gtin]] = None
+    images: Optional[List[Image]] = None
+    main_image: Optional[Image] = None
+    mpn: Optional[str] = None
+    name: Optional[str] = None
+    price: Optional[str] = None
+    regular_price: Optional[str] = None
```
Gallaecio, line 34: Should we consider `Decimal` for price fields?

zyte_common_items/items.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@export
+@attrs.define(slots=True, kw_only=True)
+class Product(_ProductBase):
+    aggregate_rating: Optional[AggregateRating] = None
+    brand: Optional[Brand] = None
+    breadcrumbs: Optional[List[Breadcrumb]] = None
+    description: Optional[str] = None
+    description_html: Optional[str] = None
```
Gallaecio, line 54: Should we make  `description_html` something else that offers selection methods?

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
       - id: black
         language_version: python3
     repo: https://github.com/ambv/black
-    rev: 22.1.0
+    rev: 22.3.0
```
Gallaecio, line 6: https://stackoverflow.com/a/71674345/939364

zyte_common_items/items.py
```diff
@@ -0,0 +1,58 @@
...
+    size: Optional[str] = None
+    sku: Optional[str] = None
+    style: Optional[str] = None
+    url: ResponseUrl
+
+
+@export
+@attrs.define(slots=True, kw_only=True)
+class ProductVariant(_ProductBase):
+    pass
```
kmike, line 44: url should be optional for ProductVariant

zyte_common_items/items.py
```diff
@@ -0,0 +1,58 @@
...
+@attrs.define(slots=True, kw_only=True)
+class Product(_ProductBase):
+    aggregate_rating: Optional[AggregateRating] = None
+    brand: Optional[Brand] = None
+    breadcrumbs: Optional[List[Breadcrumb]] = None
+    description: Optional[str] = None
+    description_html: Optional[str] = None
+    features: Optional[List[str]] = None
+    metadata: Metadata
+    product_id: Optional[str] = None
```
kmike, line 57: productId should be available for variants (it's back)

zyte_common_items/base.py
kmike: I think we're using floats, unless we really want to preserve the precision, like in price. For aggregate rating value it doesn't matter, we should use float here.

Gallaecio:
I have updated names and types to match Unified Schema as @kmike suggested elsewhere. I wonder about the new classes in `base.py`, though:

- Would they be problematic in the same sense as old names and types could be? Should we replace them by dictionaries?
- Should they have a mechanism like that of `Item` to avoid data drop if they get additional fields in the future?

zyte_common_items/base.py
```diff
@@ -118,3 +115,45 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+
+@_export_attrs()
+class Brand:
+    name: str
+
+
+@_export_attrs()
+class Breadcrumb:
+    name: str
+    link: str
```
kmike, line 141: I think that's fine to have one of the fields absent in Breadcrumbs, i.e. probably both should be declared as Optional

zyte_common_items/base.py
```diff
@@ -118,3 +115,45 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+
+@_export_attrs()
+class Gtin:
+    type: str
+    value: str
+
+
+@_export_attrs(kw_only=True)
+class Image:
+    data_url: Optional[str] = None
```
kmike, line 152:
```suggestion
    dataUrl: Optional[str] = None
```

zyte_common_items/base.py
```diff
@@ -118,3 +115,45 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+    url: Optional[str] = None
+
+
+@_export_attrs(kw_only=True)
+class Metadata:
+    date_downloaded: str
```
kmike, line 158:
```suggestion
    dateDownloaded: str
```

zyte_common_items/items.py
```diff
@@ -0,0 +1,52 @@
...
+
+@_export_attrs(kw_only=True)
+class Product(_ProductBase):
+    aggregateRating: Optional[AggregateRating] = None
+    brand: Optional[Brand] = None
+    breadcrumbs: Optional[List[Breadcrumb]] = None
+    description: Optional[str] = None
+    descriptionHtml: Optional[str] = None
+    features: Optional[List[str]] = None
+    metadata: Metadata
```
kmike, line 50: as both fields in metadata are optional, I think we can declare metadata as Optional as well

zyte_common_items/util.py
```diff
@@ -2,7 +2,7 @@
 from typing import Any, Callable, Dict, List, Optional, Tuple, Type
 from weakref import WeakKeyDictionary
 
-import attr
+import attrs
```
kmike, line 5: It might be good to define minimum attrs version in setup.py, see https://www.attrs.org/en/stable/names.html#tl-dr - probably it should be 21.3.0

zyte_common_items/base.py
```diff
@@ -118,3 +115,45 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+    probability: Optional[float] = None
```
kmike, line 159: I think the latest format is to have probability required, not optional. We can set it to 1.0 by default, so that users don't need to worry about it. It would also mean we might have a default for metadata field.

zyte_common_items/base.py
kmike: No, it's the opposite :) it may be required by the guidelines, but likely it won't be available in the API for some time.

zyte_common_items/base.py
```diff
@@ -118,3 +115,45 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+class AdditionalProperty:
+    name: str
+    value: str
+
+
+@_export_attrs(kw_only=True)
+class AggregateRating:
+    best_rating: Optional[float] = None
+    rating_value: Optional[float] = None
+    review_count: Optional[int] = None
```
kmike, line 130:
```suggestion
    bestRating: Optional[float] = None
    ratingValue: Optional[float] = None
    reviewCount: Optional[int] = None
```

kmike:
@Gallaecio 

> Would they be problematic in the same sense as old names and types could be? Should we replace them by dictionaries?

I hope no, but it'd be good to have tests for it.

> Should they have a mechanism like that of Item to avoid data drop if they get additional fields in the future?

Yes, a good catch.

tests/test_items.py
```diff
@@ -0,0 +1,190 @@
...
+                "mpn": None,
+                "name": None,
+                "price": None,
+                "productId": None,
+                "regularPrice": None,
+                "size": None,
+                "sku": None,
+                "style": None,
+                "url": None,
+            },
```
Gallaecio, line 176: Is this OK for an empty variant? Or should we make it so that unset fields are excluded from serialization somehow?

If we do _not_ exclude fields, should we have a value other than `None` as default for some fields (e.g. `[]` for lists)?

zyte_common_items/base.py
```diff
@@ -118,3 +120,52 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+
+@export
+@attrs.define(slots=True)
+class AdditionalProperty:
+    name: str
+    value: str
+
+
+@export
+@attrs.define(slots=True, kw_only=True)
```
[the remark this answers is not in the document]
Gallaecio: [I did it](https://github.com/zytedata/zyte-common-items/pull/2/commits/9f3992b4f31132409710827a261a8c2127a9e2bb), and it passed regular tests, but not mypy, so I [reverted](https://github.com/zytedata/zyte-common-items/pull/2/commits/a3511d2431cc418c543613bde3c42bc859e9079f). First I tried a few things with `functools.wraps`, but nothing I tried worked for mypy. I am sure it _can_ be done, but I do not know the right way to do it. Please, feel free to do it yourself if you feel motivated :slightly_smiling_face:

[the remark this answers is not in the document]
BurnzZ: Let's tackle it outside this PR to prevent it from being a blocker. I've created a separate issue in https://github.com/zytedata/zyte-common-items/issues/5.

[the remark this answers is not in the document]
kmike: We can also drop the `export` decorator :)

tests/test_items.py
kmike: We should exclude the following from export:

* None values
* empty lists
* empty dictionaries

zyte_common_items/base.py
```diff
@@ -91,30 +93,76 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+@attrs.define(slots=True)
+class Gtin(Item):
+    type: str
+    value: str
+
+
+@export
+@attrs.define(slots=True, kw_only=True)
+class Image(Item):
+    dataUrl: Optional[str] = None
```
kmike, line 160: we decided to drop dataUrl fields from everywhere. 
"url" field shouldn't contain data urls (they need to be filtered out); validation is out of scope for this PR, we can handle it separately.

zyte_common_items/util.py
```diff
@@ -8,6 +9,18 @@
 CLASS_ATTRS: WeakKeyDictionary = WeakKeyDictionary()
 
 
+def _remove_empty_values(obj):
+    if isinstance(obj, dict):
+        for key in list(obj):
+            if obj[key] in (None, [], {}):
```
BurnzZ, line 15: Is an empty tuple also a possibility?

zyte_common_items/base.py
```diff
@@ -16,7 +16,13 @@
...
+import attrs
+from itemadapter import ItemAdapter
+
+from zyte_common_items.util import (
+    _remove_empty_values,
+    split_in_unknown_and_known_fields,
+)
 
 
 def is_data_container(cls_or_obj):
```
BurnzZ, line 28: Minor and optional: we could probably move this to `util.py`.

tests/test_items_base.py
```diff
@@ -61,3 +78,18 @@ def test_item_from_dict_value_error():
...
+
+
+def test_metadata_optional_fields():
+    Metadata(
+        dateDownloaded="2022-12-31T13:01:54Z",
+    )
+    Metadata(
+        probability=1.0,
+    )
+    Metadata()
```
BurnzZ, line 95: minor: this is the same test case with line 84

Gallaecio: Nice catch. In fact, the whole test function can probably be removed, since `Metadata()` working implies that the other two fields can be missing.

zyte_common_items/base.py
Gallaecio: It depends on the `Item` class defined below, so it would not be trivial.

Gallaecio: We could redefine it as a class method of `Item`, e.g. `is_instance_or_subclass`.

BurnzZ: Let's leave it as-is for now.

BurnzZ (approved):
LGTM @Gallaecio ! +1 to merge after the schema is finalized.

zyte_common_items/base.py
```diff
@@ -91,30 +102,68 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+    url: Optional[str] = None
+
+
+@attrs.define(slots=True)
+class Gtin(Item):
+    type: str
+    value: str
+
+
+@attrs.define(slots=True, kw_only=True)
```
kmike, line 161: What do you think about removing kw_only for Image?

zyte_common_items/base.py
```diff
@@ -91,30 +102,68 @@ def _apply_field_types_to_sub_fields(cls, item: Optional[Dict]):
...
+class Image(Item):
+    url: Optional[str]
```
kmike, line 163: shouldn't url be required, as it's the only field?

zyte_common_items/util.py
```diff
@@ -8,6 +10,19 @@
...
 
+def _remove_empty_values(obj):
+    if isinstance(obj, dict):
+        for key in list(obj):
+            value = obj[key]
+            if value is None or (isinstance(value, (dict, Sequence)) and not value):
+                del obj[key]
+            else:
+                _remove_empty_values(obj[key])
+    elif isinstance(obj, list):
```
kmike, line 21: that's another place where we might support tuples

zyte_common_items/base.py
```diff
@@ -36,12 +42,17 @@ class _ItemBase:
...
+@attrs.define(slots=True)
 class Item(_ItemBase):
     def __attrs_post_init__(self):
         self._unknown_fields_dict = {}
 
+    def asdict(self):
+        adapter = ItemAdapter(self)
+        result = adapter.asdict()
+        _remove_empty_values(result)
+        return result
```
kmike, line 54: In autoextract-poet there was a custom adapter (https://github.com/scrapinghub/autoextract-poet/blob/master/autoextract_poet/adapters.py); I think it was used to preserve unknown fields. 

Questions:

1. Do we need to copy the approach, or are unknown fields handled in some other way? 
2. If we have a custom adapter, would it make sense to have _remove_empty_values somewhere in the adapter? I'm not sure it's possible to override `adapter.asdict()` properly, it might require changes to itemadapter.

tests/test_items.py
```diff
@@ -0,0 +1,208 @@
...
+    serialized_product = product.asdict()
+    assert serialized_product == {
+        "url": "https://example.com/?product=product22",
+        "variants": ({},),
+    }
+
+
+def test_product_asdict_tuple_empty():
+    product = Product(**_PRODUCT_MIN_KWARGS, variants=tuple())
+    serialized_product = product.asdict()
```
kmike, line 160: I think the goal is to make `ItemAdapter(product).asdict()` working, so that these items work properly as Scrapy items.

tests/test_items.py
```diff
@@ -0,0 +1,208 @@
...
+    assert serialized_product == {
+        "url": "https://example.com/?product=product22",
+    }
+
+
+def test_product_unknown_input():
```
kmike, line 166: we should make sure that unknown_values are preserved when exporting (and maybe also accessing) data via ItemAdapter.

zyte_common_items/util.py
```diff
@@ -48,6 +63,11 @@ def export(fn):
     ``__all__`` variable in the module. Useful to control
     what is imported when ``import * from <module>`` is used.
     """
+    warn(
+        "The export decorator is deprecated",
+        DeprecationWarning,
+        stacklevel=2,
+    )
```
kmike, line 70: at this stage I think it's fine to remove it, without deprecation

tests/test_adapter.py
```diff
@@ -0,0 +1,370 @@
...
+        "mpn": "HSC0424PP",
+        "name": "White two-door refrigerator",
+        "price": "9999.99",
+        "productId": "A123DK9823",
+        "regularPrice": "11999.99",
+        "size": "XL",
+        "sku": "A123DK9823",
+        "style": "polka dots",
+        "url": "https://example.com/?product=product22",
+        "variants": [{}],
```
kmike, line 74: That's an interesting case! On a first sight, I'd expect it to be dropped from the output.

Gallaecio:
> I think it'd be great to add more documentation to the field and class definitions. Would it make sense to do it in the same PR, or to handle it separately?

I am OK with doing it here.

I do wonder, though: do we do it in docstrings only, also in the README [à la itemadapter](https://github.com/scrapy/itemadapter#api-reference), or go for Sphinx already?

kmike:
> I do wonder, though: do we do it in docstrings only, also in the README [à la itemadapter](https://github.com/scrapy/itemadapter#api-reference), or go for Sphinx already?

Add docstrings here, add Sphinx in a separate PR, don't create a huge README?

zyte_common_items/adapter.py
```diff
@@ -0,0 +1,92 @@
...
+            raise KeyError(
+                f"Object of type {self.item.__class__.__name__} does not contain a field with name {field_name}"
+            )
+
+    def __iter__(self) -> Iterator:
+        fields = [attr for attr in self._fields_dict if not _is_empty(getattr(self.item, attr))]
+        fields.extend(
+            attr
+            for attr in self.item._unknown_fields_dict
+            if (attr not in fields and not _is_empty(self.item._unknown_fields_dict[attr]))
```
kmike, line 90: Why is `attr not in fields` check needed, is it an additional safety measure? I notice it because of a possible performance impact, as a list lookup.

It's hard to imagine a scenario where `fields` list is large, so the overhead is small, but it still may worth it to make it a set lookup. It's always a worst-case lookup here, the whole fields need to be iterated.

Actually, I'm also not sure how fields.extend handles the extension - if `fields` is changed on each iteration, the `not in fields` lookup makes the loop quadratic.

Some numbers, to demonstrate that my comment is not very important :) 
```py
In [1]: lst = list("abcdefsldjfgh")

In [2]: lst
Out[2]: ['a', 'b', 'c', 'd', 'e', 'f', 's', 'l', 'd', 'j', 'f', 'g', 'h']

In [3]: lst_set = set(lst)

In [4]: %timeit 'A' in lst
238 ns ± 0.708 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: %timeit 'A' in lst_set
30.7 ns ± 0.0337 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit set(lst)
265 ns ± 0.121 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

zyte_common_items/items.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+@attrs.define(slots=True, kw_only=True)
+class ProductVariant(_ProductBase):
+    """:class:`Product` variant.
+
+    See :attr:`Product.variants`.
+    """
+
+    #: Main URL from which the product variant data was extracted.
```
kmike, line 158: It should be an URL where the complete data about a variant could be extracted, not an URL from which the data in ProductVariant record was extracted

zyte_common_items/components.py
```diff
@@ -0,0 +1,133 @@
...
+    #: The probability (0 for 0%, 1 for 100%) that the webpage features the
+    #: requested data type.
+    #:
+    #: For example, if the extraction of a product from a given URL is
+    #: requested, and that URL points to the webpage of a product with complete
+    #: certainty, the value should be `1`. If with complete certainty the
+    #: webpage features a job listing instead of a product, the value should be
+    #: `0`. When there is no complete certainty, the value could be anything in
+    #: between (e.g. `0.96`).
+    probability: Optional[float] = None
```
kmike, line 133: We could set probability default to 1.0, and also create a default factory for Metadata in Product; this would allow to get it in the output, if that's desired. But let's handle it separately.

zyte_common_items/items.py
Gallaecio: How is it different from `canonicalUrl`?

kmike:
Thanks @Gallaecio! 🚀 🚀

PR #3, opened 2022-06-22
Run static checks only once

Gallaecio:
Split out of https://github.com/zytedata/zyte-common-items/pull/2

https://github.com/zytedata/zyte-common-items/pull/4 should address the test failure.


BurnzZ (approved):
Nice catch! 🙌

PR #6, opened 2022-07-20
Item.from_page_object

kmike:
This is an alternative to https://github.com/scrapinghub/web-poet/pull/53.

It allows to implement page objects this way:

```py
@attrs.define
class MyProductPage(ItemPage):
    response: HttpResponse

    def url(self) -> str:
        return self.response.url

    def price(self) -> str:
        return self.response.css(".price").get()

    async def currency(self) -> str:
        return "$"

    async def to_item(self) -> Product:
        item = await Product.from_page_object(self)
        # further processing, if needed
        return item
```

With a mixin (not implemented), it could look like this:

```py
@attrs.define
class MyProductPage(ItemPage, AutoMixin[Product]):
    response: HttpResponse

    def url(self) -> str:
        return self.response.url

    def price(self) -> str:
        return self.response.css(".price").get()

    async def currency(self) -> str:
        return "$"
```

### DIFF
--- a/zyte_common_items/base.py
+++ b/zyte_common_items/base.py
@@ -18,7 +18,7 @@ from typing import Dict, List, Optional, Union
 
 import attrs
 
-from zyte_common_items.util import split_in_unknown_and_known_fields
+from zyte_common_items.util import split_in_unknown_and_known_fields, maybe_await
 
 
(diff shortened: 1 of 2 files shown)


kmike:
Closing it, as the alternative approach seems to be better.

PR #1, opened 2022-03-08
Add items for these schemas: Zyte Data API, Unified Schema

BurnzZ:
## Progress

- [x] Zyte Data API Schema Items
  - [x] tests
- [x] Unified Schema Items
  - [ ] tests
- [ ] Changelog
- [ ] Documentation

PR #7, opened 2022-07-26
Add ProductList

Gallaecio:
To avoid some repetition I needed multiple-inheritance, which does not work with slotted attrs (3rd bullet point of https://www.attrs.org/en/stable/glossary.html#term-slotted-classes).

I decided to not use slots, since I don’t think the benefits are worth the drawbacks here. But if you think keeping slots is worth it, I can adapt the implementation accordingly.

In fact, I would understand if you think it would be best not to go so hard on avoiding code repetition here, and instead redefine properties in as many classes as needed. But even then, I think disabling slots may be best to give more freedom to people subclassing us.

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -83,7 +83,7 @@ _PRODUCT_ALL_KWARGS = {
     "variants": [ProductVariant()],
 }
 _PRODUCT_LIST_MIN_KWARGS = {
-    "url": "https://example.com/?product=product22",
+    "url": "https://example.com/swiss-watches?sort=new-first",
 }
 _PRODUCT_LIST_ALL_KWARGS = {
     **_PRODUCT_LIST_MIN_KWARGS,
@@ -91,14 +91,17 @@ _PRODUCT_LIST_ALL_KWARGS = {
         Breadcrumb(name="Level 1", url="http://example.com/level1"),
         Breadcrumb(name="Level 2", url="http://example.com/level1/level2"),
     ],
-    "canonicalUrl": "https://example.com/product22",
+    "canonicalUrl": "https://example.com/swiss-watches",
     "categoryName": "Swiss Watches",
     "metadata": Metadata(
         dateDownloaded="2022-12-31T13:01:54Z",
         probability=1.0,
     ),
     "pageNumber": 1,
-    "paginationNext": Link(text="foo", url="https://example.com"),
+    "paginationNext": Link(
+        text="foo",
+        url="https://example.com/swiss-watches?sort=new-first&page=2",
+    ),
     "products": [ProductListItem()],
 }
 
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -12,12 +12,12 @@ from zyte_common_items import (
     Link,
     Metadata,
     Product,
+    ProductFromList,
     ProductList,
-    ProductListItem,
     ProductVariant,
 )
 
-_PRODUCT_LIST_ITEM_ALL_KWARGS = {
+_PRODUCT_FROM_LIST_ALL_KWARGS = {
     "currency": "USD",
     "currencyRaw": "$",
     "mainImage": Image("http://example.com/image1.png"),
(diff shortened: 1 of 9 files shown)


kmike:
> In fact, I would understand if you think it would be best not to go so hard on avoiding code repetition here, and instead redefine properties in as many classes as needed. But even then, I think disabling slots may be best to give more freedom to people subclassing us.

I think some repetition is fine, and probably readability benefits worth having some code duplication, instead of going with mixins & hierarchies. Not having slots is fine.

BurnzZ:
> I think some repetition is fine, and probably readability benefits worth having some code duplication, instead of going with mixins & hierarchies.

+1 on this.

After toying with https://github.com/zytedata/zyte-common-items/pull/1/files, I realized that it's much easier to read (even if code is repeated) if all of the fields are placed and easily viewable inside the item class.

I find it difficult to trace each of the parent classes to see what are the other fields that might belong to a particular item. Another way to think about it would be the case when developers want to populate a given item with values. If the fields are all available in the class (without inheritance) then they could use it as a direct checklist.

As an exercise, suppose I want to create `ProductList` and populate all the available fields. First, I'd go to the class definition of [ProductList](https://github.com/zytedata/zyte-common-items/pull/7/files#diff-9247aa1d02420dcd380f7568ddee4580ab6fbe03d5ce871ab8bb72f4088b2a07R309). I see that the fields `categoryName` and `products` are present. Moreover, I see that it's subclassing from the `_Page` and `_Paginated` classes. So I'd go to the [_Page](https://github.com/zytedata/zyte-common-items/pull/7/files#diff-9247aa1d02420dcd380f7568ddee4580ab6fbe03d5ce871ab8bb72f4088b2a07R178) and [_Paginated](https://github.com/zytedata/zyte-common-items/pull/7/files#diff-9247aa1d02420dcd380f7568ddee4580ab6fbe03d5ce871ab8bb72f4088b2a07R295) definition. I take note of all of the fields from each one.

I think it would be easier if `ProductList` listed out all of the fields from the get go.

tests/test_items.py
```diff
@@ -65,6 +82,25 @@
     ),
     "variants": [ProductVariant()],
 }
+_PRODUCT_LIST_MIN_KWARGS = {
+    "url": "https://example.com/?product=product22",
```
kmike, line 86: a nitpick: something which looks like a product list url could be a better example, e.g. example.com/watches/swiss (to match categoryName)

kmike:
> But even then, I think disabling slots may be best to give more freedom to people subclassing us.

Ouch, I now recall why we're using slots - it's to ensure that an assignment to an incorrect item field raises an AttributeError.

Gallaecio:
Slots prevent multiple inheritance. If we are OK with disallowing that, I see no problem using them.

kmike (approved):
👍

zyte_common_items/items.py
```diff
@@ -252,3 +376,123 @@ class Product(_ProductBase):
...
+    #:
+    #: The order of the products reflects their position on the rendered page.
+    #: Product order is top-to-bottom, and left-to-right or right-to-left
+    #: depending on the webpage locale.
+    products: Optional[List[ProductListItem]] = None
+
+    #: Main URL from which the data has been extracted.
+    #:
+    #: See also ``canonicalUrl``.
+    url: str
```
BurnzZ, line 498: Should we add the `metadata` field as well?

Gallaecio: It is a few fields above (fields are in alphabetical order).

BurnzZ: My bad, I totally missed it!

BurnzZ (approved):
LGTM @Gallaecio ! 🚀

zyte_common_items/components.py
```diff
@@ -110,6 +110,17 @@ class Image(Item):
     url: str
 
 
+@attrs.define(slots=True, kw_only=True)
```
kmike, line 113: slots=True is redundant here

kmike:
Thanks @Gallaecio!

zyte_common_items/components.py
kmike: it's handled in another PR, so merging :)

PR #8, opened 2022-07-26
Use Sphinx for documentation

Gallaecio:
I hope I did not go too far :)


setup.py
```diff
@@ -14,8 +19,10 @@
             "tests",
         ]
     ),
+    include_package_data=True,
     install_requires=[
         "attrs>=21.3.0",
+        "itemadapter>=0.2.0",
```
Gallaecio, line 25: Is this OK? I think it may have been a mistake on my part not to include it when adding the adapter.

tests/typing.py
```diff
@@ -10,28 +10,28 @@
...
 def optional_type(cls):
-    """
-    If cls is Optional[X] it returns X else returns None
+    """If cls is Optional[X] it returns X else returns None.
 
     >>> optional_type(Optional[str])
     <class 'str'>
     >>> optional_type(Union[str, int])
     >>> optional_type(str)
+
```
Gallaecio, line 19: Adding this extra line at the end makes doctests work for Sybil.

setup.py
Gallaecio: Note to self: remove from `tox.ini` if this is OK.

kmike: yes, it should be a dependency, nice catch @Gallaecio!

README.rst
```diff
@@ -18,146 +18,21 @@ zyte-common-items
...
-    }
-    article_list = zda.ArticleList.from_dict(data)    
-
-The amazing feature here is that the ``from_dict()`` method creates the correct sub-item class instance from the main item (see ``PaginationLink`` and ``ArticleFromList`` below).
-
-.. code-block:: python
-
-    print(article_list.url)
-    # https://example.com/articles/fashion    
+It can be used, for example, to load item data from `Zyte Automatic
```
kmike, line 26: This is not true, because the data format is different. We could add this note back when we launch Zyte API with extraction.

docs/contributing.rst
```diff
@@ -0,0 +1,19 @@
+============
+Contributing
```
kmike, line 2: We should probably mention the repository and issue tracker in this section.

docs/index.rst
```diff
@@ -0,0 +1,15 @@
+=========================================
+zyte-common-items |version| documentation
+=========================================
+
+.. include:: ../README.rst
+   :start-after: .. description starts
+   :end-before: .. description ends
```
kmike, line 7: TIL

docs/usage.rst
```diff
@@ -0,0 +1,93 @@
...
+'Product 4'
+
+This can be especially useful if you're processing lots of items from an API,
+file, database, etc.
+
+
+Handling unknown fields
+=======================
+
+Fields that are not defined in :ref:`items <items>` **won't cause an error**.
```
kmike, line 66: They won't cause an error if from_dict or from_list is used, but I think they still cause an error if `__init__` is used (and probably that's the right behavior).

docs/usage.rst
```diff
@@ -0,0 +1,93 @@
...
+...     'unknown_field': True,
+... }
+>>> product = Product.from_dict(data)
+>>> product._unknown_fields_dict
+{'unknown_field': True}
+
+This allows compatibility with future field changes in the input data, which
+could cause backwards incompatibility issues.
+
+However, if a known field has an incorrect data structure,
```
kmike, line 82: It may read as we do type validation, but I'm not sure that's the case. Error only happens in some cases, as a side effect.

tox.ini
```diff
@@ -6,12 +6,23 @@ deps =
...
     pytest-cov
     itemadapter
-
+    sybil
+setenv =
+    PY_IGNORE_IMPORTMISMATCH=1
 commands =
     py.test \
         --cov-report=term-missing:skip-covered --cov-report=html --cov-report= --cov=zyte_common_items \
-        --doctest-modules \
```
kmike, line 13: Just curious, what's the advantage of using sybil instead of --doctest-modules?

kmike:
Nice, looks like a great foundation @Gallaecio!

tox.ini
Gallaecio: I expected `--doctest-modules` to work right away, but it did not catch reStructuredText files after including `docs` in the command targets, so I started trying to find out why, reached https://docs.pytest.org/en/7.1.x/how-to/doctest.html#alternatives, failed to get the first suggested package, but managed to get Sybil working.

However, I did not read that page from the beginning, and I thought `--doctest-modules` could not support rst because of how they highlight the fact that the first suggested alternative supports rst. But I never tried `--doctest-glob="*.rst"`, featured at the very beginning of the page.

I do know that Sybil has some nice features, allowing e.g. to write hidden doctest content, not visible in the documentation but executed during tests. But I will try `--doctest-glob="*.rst"`, and if it works for us now, I think it best to use it and keep things simple.

docs/usage.rst
```diff
@@ -0,0 +1,137 @@
...
+
+
+Defining custom items
+=====================
+
+You can subclass :class:`~zyte_common_items.base.Item` or any :ref:`item
+subclass <items>` to define your own item.
+
+:class:`~zyte_common_items.base.Item` is a `slotted attrs class`_, and
+subclasses should also be slotted attrs classes. For example:
```
kmike, line 127: If I'm not mistaken, there is no strict requirement for subclasses to be slotted. Or is it? If a subclass is not slotted, it just means that assigning a non-existing attribute doesn't raise an error; it's an important gotcha, but that's not specific to items.

docs/usage.rst
```diff
@@ -0,0 +1,137 @@
...
+>>> product = Product(url="https://example.com")
+>>> product.foo = "bar"
+Traceback (most recent call last):
+...
+AttributeError: 'Product' object has no attribute 'foo'
+
+However, when using :meth:`~zyte_common_items.base.Item.from_dict` and
+:meth:`~zyte_common_items.base.Item.from_list`, unknown fields assigned to
+items and components **won't cause an error**. Instead, they are placed inside
+the :attr:`~zyte_common_items.base.Item._unknown_fields_dict` attribute:
```
kmike, line 83: It'd be nice to mention ItemAdapter here - in most cases, users shouldn't use `_unknown_fields_dict` directly.

BurnzZ (approved):
Well done @Gallaecio ! 👏

docs/usage.rst
Gallaecio: https://github.com/zytedata/zyte-common-items/pull/8/commits/6bd0e7fa2d0fc166124dddd7fe83d506cd23bad1

I did use the adapter class directly instead of `itemadapter.ItemAdapter` because getting doctest to pass with the later would require either making the doctest too verbose or switching back to Sybil to have hidden setup code for it. But to be honest, I think it makes sense for ZyteItemAdapter to be used directly in scenarios like this.

kmike:
Thanks @Gallaecio!

PR #10, opened 2022-08-15
run mypy on untyped test functions; fix typing in tests

BurnzZ:
Fortifies the type checks in mypy.

### DIFF
--- a/tests/test_adapter.py
+++ b/tests/test_adapter.py
@@ -25,7 +25,7 @@ def configured_adapter():
 
 
 def test_asdict_all_fields():
-    product = Product(**_PRODUCT_ALL_KWARGS)  # type: ignore[arg-type]
+    product = Product(**_PRODUCT_ALL_KWARGS)
     with configured_adapter():
         adapter = ItemAdapter(product)
         actual_dict = adapter.asdict()
@@ -77,7 +77,7 @@ def test_asdict_all_fields():
 
 
 def test_product_asdict_min():
-    product = Product(**_PRODUCT_MIN_KWARGS)  # type: ignore[arg-type]
+    product = Product(**_PRODUCT_MIN_KWARGS)
     with configured_adapter():
         adapter = ItemAdapter(product)
         actual_dict = adapter.asdict()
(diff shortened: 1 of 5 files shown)


tests/test_adapter.py
```diff
@@ -25,7 +25,7 @@ def configured_adapter():
 
 
 def test_asdict_all_fields():
-    product = Product(**_PRODUCT_ALL_KWARGS)
+    product = Product(**_PRODUCT_ALL_KWARGS)  # type: ignore[arg-type]
```
BurnzZ, line 28: Not quite satisfied in adding `# type: ignore[arg-type]` to suppress the mypy errors here. Nonetheless, it's needed because mypy is strict in the order of the constructor arguments and `_PRODUCT_ALL_KWARGS` doesn't follow the order.

kmike: > Not quite satisfied in adding # type: ignore[arg-type] to suppress the mypy errors here. Nonetheless, it's needed because mypy is strict in the order of the constructor arguments and _PRODUCT_ALL_KWARGS doesn't follow the order.

What do you mean by "order" @BurnzZ?

BurnzZ: My bad @kmike I got confused with **mypy**'s errors like:
```
error: Argument 1 to "Foo" has incompatible type "**Dict[str, object]"; expected "int"
error: Argument 1 to "Foo" has incompatible type "**Dict[str, object]"; expected "str"
```

I fixed this by correctly annotating the definitions like `_PRODUCT_ALL_KWARGS` instead. https://github.com/zytedata/zyte-common-items/pull/10/commits/36bd3ac8548144f81d3b9b21cf92b1c9922784d9

tox.ini
```diff
@@ -28,4 +28,9 @@ basepython = python3.10
 deps =
     mypy==0.910
 
-commands = mypy --ignore-missing-imports --no-warn-no-return zyte_common_items tests
+commands = mypy \
+    --check-untyped-defs \
```
kmike, line 32: `--check-untyped-defs` may cause a bit higher maintenance overhead, not sure. But let's try it; we can revert later if it's a burden.

kmike:
Thanks @BurnzZ!

PR #11, opened 2022-08-16
Support for web-poet URL classes

BurnzZ:
Built on top of https://github.com/zytedata/zyte-common-items/pull/10 which contains the fixes for running mypy properly on tests.

----

This PR enables auto-conversion of `web-poet` URL classes (i.e. `RequestURL` and `ResponseUrl`) into fields that could use such URL classes (e.g. `mainImage`, `url`, etc.).

## TODO:

- [ ] Update `setup.py` to point to the new `web-poet` release conforming to PEP 561.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,2 @@
 Sphinx==5.1.0
 sphinx-rtd-theme==1.0.0
-web-poet
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 Sphinx==5.1.0
 sphinx-rtd-theme==1.0.0
+web-poet
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,8 +1,0 @@
-[mypy]
-check_untyped_defs = True
-ignore_missing_imports = True
-no_warn_no_return = True
-exclude = (?x)(
-    test_mypy\.py$
-    | test_conversion\.py$
-  )
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,8 @@
+[mypy]
+check_untyped_defs = True
+ignore_missing_imports = True
+no_warn_no_return = True
+exclude = (?x)(
+    test_mypy\.py$
+    | test_conversion\.py$
+  )
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
         "typing-extensions>=4.3.0;python_version<'3.9'",
-        "web-poet>=0.4.0",
+        "web-poet @ git+ssh://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
     ],
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup(
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
         "typing-extensions>=4.3.0",
+        "web-poet>=0.4.0",
     ],
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
-        "typing-extensions>=4.3.0;python_version<'3.9'",
+        "typing-extensions>=4.3.0;python_version<'3.10'",
         "web-poet @ git+https://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
     ],
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,6 @@ setup(
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
-        "typing-extensions>=4.3.0;python_version<'3.10'",
         "web-poet>=0.4.0",
     ],
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,7 @@ setup(
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
+        "typing-extensions>=4.3.0",
     ],
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ setup(
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
-        "typing-extensions>=4.3.0",
+        "typing-extensions>=4.3.0;python_version<'3.9'",
         "web-poet>=0.4.0",
     ],
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
         "typing-extensions>=4.3.0;python_version<'3.10'",
-        "web-poet @ git+https://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
+        "web-poet>=0.4.0",
     ],
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
         "typing-extensions>=4.3.0;python_version<'3.9'",
-        "web-poet @ git+ssh://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
+        "web-poet @ git+https://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
     ],
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/tests/test_mypy.py
+++ b/tests/test_mypy.py
@@ -34,85 +34,154 @@ URL = "https://www.some-page/"
 RESPONSE = HttpResponse(URL, b"")
 
 
+# NOTES: The following test cases below for assigning ``ResponseUrl`` instances
+# are marked with "xfail" due to the following known mypy bugs:
+#   - https://github.com/python/mypy/issues/10187
+#   - https://github.com/python/mypy/issues/5313
+#   - https://github.com/python-attrs/attrs/issues/889
+# Currently, the said bugs causes mypy to raise the following error:
+#   'Incompatible types in assignment (expression has type "ResponseUrl",
+#   variable has type "Optional[str]")'
+# FIXME: Unmark these from "xfail" once the upstream mypy issues are solved.
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_image_url():
+    img = Image(url=RESPONSE.url)
+    img.url = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_breadcrumb_url():
+    bc = Breadcrumb(url=RESPONSE.url)
+    bc.url = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_link_url():
+    link = Link(url=RESPONSE.url)
+    link.url = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_product_url():
+    product_1 = Product(url=RESPONSE.url)
+    product_1.url = RESPONSE.url
+
+    product_2 = Product(url=RESPONSE.url)
+    product_2.canonicalUrl = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_product_list_url():
+    product_list_1 = ProductList(url=RESPONSE.url)
+    product_list_1.url = RESPONSE.url
+
+    product_list_2 = ProductList(url=RESPONSE.url)
+    product_list_2.canonicalUrl = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_product_from_list_url():
+    product_from_list_1 = ProductFromList(url=RESPONSE.url)
+    product_from_list_1.url = RESPONSE.url
+
+
+@pytest.mark.mypy_testing
+@pytest.mark.xfail
+def test_assignment_product_variant_url():
+    product_variant_1 = ProductVariant(url=RESPONSE.url)
+    product_variant_1.url = RESPONSE.url
+
+    product_variant_2 = ProductVariant(url=RESPONSE.url)
+    product_variant_2.canonicalUrl = RESPONSE.url
+
+
 @pytest.mark.mypy_testing
 def test_assignment_image():
     img_1 = Image(url=RESPONSE.url)
-    img_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "str")
+    img_1.url = str(RESPONSE.url)
 
     img_2 = Image(url=RESPONSE.url)
-    img_2.url = str(RESPONSE.url)
+    img_2.url = "https://www.example.com"
 
 
 @pytest.mark.mypy_testing
 def test_assignment_breadcrumb():
     bc_1 = Breadcrumb(url=RESPONSE.url)
-    bc_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "Optional[str]")
+    bc_1.url = str(RESPONSE.url)
 
     bc_2 = Breadcrumb(url=RESPONSE.url)
-    bc_2.url = str(RESPONSE.url)
+    bc_2.url = "https://www.example.com"
 
 
 @pytest.mark.mypy_testing
 def test_assignment_link():
     link_1 = Link(url=RESPONSE.url)
-    link_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "Optional[str]")
+    link_1.url = str(RESPONSE.url)
 
     link_2 = Link(url=RESPONSE.url)
-    link_2.url = str(RESPONSE.url)
+    link_2.url = "https://www.example.com"
 
 
 @pytest.mark.mypy_testing
 def test_assignment_product():
     product_1 = Product(url=RESPONSE.url)
-    product_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "str")
+    product_1.url = str(RESPONSE.url)
 
     product_2 = Product(url=RESPONSE.url)
-    product_2.url = str(RESPONSE.url)
+    product_2.url = "https://www.example.com"
 
     product_3 = Product(url=RESPONSE.url)
-    product_3.canonicalUrl = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "Optional[str]")
+    product_3.canonicalUrl = str(RESPONSE.url)
 
     product_4 = Product(url=RESPONSE.url)
-    product_4.canonicalUrl = str(RESPONSE.url)
+    product_4.canonicalUrl = "https://www.example.com"
 
 
 @pytest.mark.mypy_testing
 def test_assignment_product_list():
     product_list_1 = ProductList(url=RESPONSE.url)
-    product_list_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "str")
+    product_list_1.url = str(RESPONSE.url)
 
     product_list_2 = ProductList(url=RESPONSE.url)
-    product_list_2.url = str(RESPONSE.url)
+    product_list_2.url = "https://www.example.com"
 
     product_list_3 = ProductList(url=RESPONSE.url)
-    product_list_3.canonicalUrl = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "Optional[str]")
+    product_list_3.canonicalUrl = str(RESPONSE.url)
 
     product_list_4 = ProductList(url=RESPONSE.url)
-    product_list_4.canonicalUrl = str(RESPONSE.url)
+    product_list_4.canonicalUrl = "https://www.example.com"
 
 
 @pytest.mark.mypy_testing
 def test_assignment_product_from_list():
     product_from_list_1 = ProductFromList(url=RESPONSE.url)
-    product_from_list_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "Optional[str]")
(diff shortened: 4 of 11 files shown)


zyte_common_items/components.py
```diff
@@ -108,7 +109,7 @@ class Image(Item):
     #: `Data URIs`_ are not allowed in this attribute.
     #:
     #: .. _Data URIs: https://en.wikipedia.org/wiki/Data_URI_scheme
-    url: str
+    url: str = attrs.field(converter=str)
```
kmike, line 112: I wonder if we should support only Url classes from web-poet, not arbitrary objects. It seems with the current implementation you can pass anything to url (e.g. a Selector instance, or a list of strings), and it won't raise a type error.

As discussed elsewhere, it'd also be good to ensure conversion happens on assignment.

BurnzZ: > it'd also be good to ensure conversion happens on assignment.
Added tests cases in https://github.com/zytedata/zyte-common-items/pull/11/commits/7e9135dbf893921012b9de4d7325e1b33fc3de33.

> I wonder if we should support only Url classes from web-poet, not arbitrary objects. It seems with the current implementation you can pass anything to url (e.g. a Selector instance, or a list of strings), and it won't raise a type error.
I've partially implemented this locally but need to add more test cases to handle some use cases.

BurnzZ: So I've pushed some changes in https://github.com/zytedata/zyte-common-items/pull/11/commits/3e53d6154ce7e5f1f9d0ec85f7768c100de3ad7f which:

- ensures that values that are not URL classes (i.e. subclasses of `web_poet.page_inputs.url._Url`) are not auto-converted. For example, `Image(url=123)` should become `Image(url="123")`.
- ensure that **mypy** raises errors when assigning non-string or non-URL values to fields. For example, doing `img.url = 123` should error out in **mypy**.

However, one thing's that lacking is ensuring that **mypy** also raises the appropriate errors when instantiating with incorrect types. For example, `Image(url=123)` should cause **mypy** raise an error. For now, I've marked the tests with `xfail` ([reference](https://github.com/zytedata/zyte-common-items/pull/11/commits/3e53d6154ce7e5f1f9d0ec85f7768c100de3ad7f#diff-901c8a5cbbdeeb7f1d4acf5ed1da5352f3f6fa461e1f1d11de2bf8a02d5f9c08R121-R164)) until **web-poet** uploads its stubs as [mentioned here](https://github.com/zytedata/zyte-common-items/pull/11/commits/3e53d6154ce7e5f1f9d0ec85f7768c100de3ad7f#diff-901c8a5cbbdeeb7f1d4acf5ed1da5352f3f6fa461e1f1d11de2bf8a02d5f9c08R107-R118).

tests/test_mypy.py
```diff
@@ -0,0 +1,164 @@
+"""This test file is ignored by mypy since we're using pytest-mypy-testing to
```
kmike, line 1: TIL

tests/test_mypy.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# NOTE: The tests below are marked to fail since the following error is
+# encountered in mypy:
+#
+#   "error: Skipping analyzing "web_poet": module is installed, but missing
+#   library stubs or py.typed marker"
+#
+# This means that mypy renders all contents (i.e. URL classes) from web-poet to
+# be of type Any. To fix this, web-poet must either be PEP 561 complient with
```
kmike, line 114: If web-poet code is annotated, we should add py.typed marker.

tox.ini
```diff
@@ -5,6 +5,8 @@ envlist = py37,py38,py39,py310,mypy
 deps =
     pytest
     pytest-cov
+    pytest-mypy-testing==0.0.11
+    web-poet
```
kmike, line 9: Is it really needed?

zyte_common_items/util.py
```diff
@@ -54,3 +60,13 @@ def get_origin(tp) -> Tuple:
     typing.get_origin(tp) is only available startingo on 3.8.
     """
     return getattr(tp, "__origin__", ())
+
+
+def is_url_obj(url: Union[str, _Url]) -> TypeGuard[_Url]:
+    return True if isinstance(url, _Url) else False
+
+
+def url_to_str(url: Union[str, _Url]) -> object:
```
kmike, line 69: Type annotations don't look right - it seems this function accepts Any, and returns object of the same type, except for URLs, which are demoted to str. 

What's the use case for returning the url as-is if it's not a string? Why not raise an error?

zyte_common_items/util.py
```diff
@@ -54,3 +60,13 @@ def get_origin(tp) -> Tuple:
     typing.get_origin(tp) is only available startingo on 3.8.
     """
     return getattr(tp, "__origin__", ())
+
+
+def is_url_obj(url: Union[str, _Url]) -> TypeGuard[_Url]:
```
kmike, line 65: TIL TypeGuard :)

tests/test_mypy.py
```diff
@@ -0,0 +1,164 @@
...
+    bc.url = str(response.url)
+    bc.url = 123  # E: Incompatible types in assignment (expression has type "int", variable has type "Optional[str]")
+
+
+@pytest.mark.mypy_testing
+def test_assignment_link():
+    link = Link(url=response.url)
+    link.url = response.url
+    link.url = str(response.url)
+    link.url = 123  # E: Incompatible types in assignment (expression has type "int", variable has type "Optional[str]")
```
kmike, line 53: This might only fail because of a previous assignment, not because of annotation of Link objects. I.e. the following code fails mypy checks

```
x = "hello"
x = 5
# Incompatible types in assignment (expression has type "int", variable has type "str")
```

docs/requirements.txt
```diff
@@ -1,2 +1,3 @@
 Sphinx==5.1.0
 sphinx-rtd-theme==1.0.0
+web-poet
```
kmike, line 3: why is it required? I'm worried if it'd break something, insyalling we-poet from pypi instead of doing it from source.

zyte_common_items/util.py
BurnzZ: >  it seems this function accepts Any

I'm afraid I don't follow since it's annotated to accept `Union[str, _Url]`.

> What's the use case for returning the url as-is if it's not a string? Why not raise an error?

This setup was intended for static type checks only. Should we also check do some runtime checks with errors during the conversion process?

docs/requirements.txt
BurnzZ: It's for importing `web_poet.page_inputs.url._Url` to ensure that the string converter only handles those.

kmike: Ah, I see. Shouldn't we add it to install_requires instead? It seems otherwise utils.py will fail with ImportError

kmike: I was confused; for some reason I was thinking this is a PR to web-poet 🤦

tests/test_mypy.py
kmike: Actually, we should do it for zyte-common-items as well :)

tests/test_mypy.py
BurnzZ: Ahhh nice catch! I used distinct variables instead to avoid this. https://github.com/zytedata/zyte-common-items/pull/11/commits/57859876b8f9eca829d202f8fd48e97a3e19f298

docs/requirements.txt
BurnzZ: Good call! moved it to `install_requires`. https://github.com/zytedata/zyte-common-items/pull/11/commits/a23757e2080ec986c707bbe482c6268d23e1aa94

tests/test_mypy.py
BurnzZ: Created an issue in `web-poet` here: https://github.com/scrapinghub/web-poet/issues/65.

Should we also handle https://peps.python.org/pep-0561/ in this PR for `zyte-common-items`?

tests/test_mypy.py
```diff
@@ -0,0 +1,212 @@
...
+    product_variant_6.canonicalUrl = (
+        123  # E: Incompatible types in assignment (expression has type "int", variable has type "Optional[str]")
+    )
+
+
+# NOTE: The tests below are marked to fail since the following error is
+# encountered in mypy:
+#
+#   "error: Skipping analyzing "web_poet": module is installed, but missing
+#   library stubs or py.typed marker"
```
BurnzZ, line 159: Here's a code snippet to demonstrate the **mypy** behavior when the stubs in `web-poet` would be available. The following code works because the URL classes are in the same script and not imported as a third-party package.

```python
from typing import Union
from typing_extensions import TypeGuard

import attrs


class Url:
    def __init__(self, url: Union[str, "Url"]):
        if not isinstance(url, (str, Url)):
            raise TypeError("err")
        self._url = str(url)

    def __str__(self) -> str:
        return self._url


class ResponseUrl(Url):
    pass


def is_url_obj(url: Union[str, Url]) -> TypeGuard[Url]:
    return True if isinstance(url, Url) else False


def url_to_str(value: Union[str, Url]) -> object:
    if is_url_obj(value):
        return str(value)
    return value


@attrs.define
class Image:
    url: str = attrs.field(converter=url_to_str)


url = ResponseUrl("123")
img = Image(url)

# error: Incompatible types in assignment (expression has type "int", variable has type "str")
img.url = 12

# error: Argument "url" to "Image" has incompatible type "int"; expected "Union[str, Url]"
f = Image(url=123)

# Second error does not pop out if this error is raised since it would render
# all classes coming from web_poet to be of type ``Any``.
# error: Skipping analyzing "web_poet": module is installed, but missing library stubs or py.typed marker

```

tests/test_mypy.py
kmike: Thanks @BurnzZ! No need to handle it in this PR; I've created https://github.com/zytedata/zyte-common-items/pull/12 and https://github.com/scrapinghub/web-poet/pull/66.

kmike: We can't release web-poet yet though, because of unfinished fields feature.

mypy.ini
```diff
@@ -0,0 +1,8 @@
+[mypy]
```
kmike, line 1: What do you think about moving this config to pyproject.toml? It'd mean one less file on a top level. See https://mypy.readthedocs.io/en/stable/config_file.html#using-a-pyproject-toml

setup.py
```diff
@@ -23,6 +23,8 @@
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
+        "typing-extensions>=4.3.0",
```
kmike, line 26: Would it make sense to install it only in Python versions which don't support TypeGuard?

zyte_common_items/components.py
```diff
@@ -67,7 +68,7 @@ class Breadcrumb(Item):
     name: Optional[str] = None
 
     #: Target URL.
-    url: Optional[str] = None
+    url: Optional[str] = attrs.field(default=None, converter=attrs.converters.optional(url_to_str), kw_only=True)
```
kmike, line 71: Such lines are getting harder to review (I need to scroll horizontally when using github UI); it's not something to fix in this PR, but I opened https://github.com/zytedata/zyte-common-items/issues/13 for a discussion.

zyte_common_items/util.py
kmike: It seems reasonable to fail with a ValueError if conversion to str can't happen.

setup.py
BurnzZ: Good call! https://github.com/zytedata/zyte-common-items/pull/11/commits/6690ed8cbc3091f7888c70dc0b4281c32a30a556

mypy.ini
BurnzZ: Good call! https://github.com/zytedata/zyte-common-items/pull/11/commits/6690ed8cbc3091f7888c70dc0b4281c32a30a556

setup.py
BurnzZ: Nice catch! https://github.com/zytedata/zyte-common-items/pull/11/commits/f5a9088017792cd8cb3041ee5e4d7f9e5e09b208

zyte_common_items/util.py
```diff
@@ -1,7 +1,13 @@
-from typing import Any, Callable, Dict, Optional, Tuple, Type
+from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
 from weakref import WeakKeyDictionary
 
 import attr
+from web_poet.page_inputs.url import _Url
+
+try:
+    from typing import TypeGuard  # for Python 3.9+
```
kmike, line 8: https://peps.python.org/pep-0647/ states it's Python 3.10 - is the feature available in 3.9 as well?

tests/test_mypy.py
BurnzZ: Thanks @kmike ! 

> We can't release web-poet yet though, because of unfinished fields feature.

That's okay, we can temporarily point to the commits for this PR to check that the tests are working alright.

zyte_common_items/util.py
BurnzZ: Gotcha, I've updated the logic here: https://github.com/zytedata/zyte-common-items/pull/11/commits/3fee61364f5f6a93598edc0bddbc7618240d6f29

zyte_common_items/components.py
BurnzZ: Agreed.

Since we're in the topic of long-lines, I'm not quite keen on the current setup of `pytest-mypy-testing` since there's no way to breakdown the lines. I've filed an issue to their repo. In addition, I've made some workarounds to prevent Black from messing with its format: https://github.com/zytedata/zyte-common-items/pull/11/commits/3fee61364f5f6a93598edc0bddbc7618240d6f29#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R3

zyte_common_items/util.py
kmike: I think type annotation now needs to be changed - the output is str, not object.

I'm also not sure about using object in a first place, it seems Any was a better fit.

zyte_common_items/util.py
BurnzZ: Nice catch! Fixed in https://github.com/zytedata/zyte-common-items/pull/11/commits/391f617e502ba98e5abbf701c2e6243d8b15f9ad

kmike: do you need to fix setup.py as well?

zyte_common_items/util.py
BurnzZ: > I think type annotation now needs to be changed - the output is str, not object

Good catch. I've updated it in https://github.com/zytedata/zyte-common-items/pull/11/commits/fbb6ed028b46dfbad6bf8b22747d2f07d087c386.

> I'm also not sure about using object in a first place, it seems Any was a better fit.

Using `Any` would "turn off" the type checker ([ref](https://docs.python.org/3/library/typing.html#the-any-type)): "Notice that no typechecking is performed when assigning a value of type [Any](https://docs.python.org/3/library/typing.html#typing.Any) to a more precise type"

There's also this discussion that I found here: https://stackoverflow.com/questions/39817081/typing-any-vs-object

zyte_common_items/util.py
BurnzZ: Ahhh missed that one. I was under the impression I was using less than or equal. 😅 . Updated in https://github.com/zytedata/zyte-common-items/pull/11/commits/ec525edea85227932086c341630d277836b51d81

zyte_common_items/util.py
kmike: This link suggests 

> That means when the type of a value is object, a type checker will reject almost all operations on it, and assigning it to a variable (or using it as a return value) of a more specialized type is a type error.

it doesn't seem right; likely we could have used a TypeVar to annotate `url`, and make a return type a Union[T, str] -but that's irrelevant now :)

zyte_common_items/util.py
kmike: for me it's not clear why were tests passing in Python 3.9

BurnzZ: For Python 3.9, importing `TypeGuard` from `typing` would error out and it'd be imported via `typing_extensions` instead.

I also realized that https://pypi.org/project/typing-extensions/ also works until 3.10.

zyte_common_items/util.py
BurnzZ: Ahhh that's a good point. I missed out on this since the code wasn't accessing any methods from the object. 👍

setup.py
```diff
@@ -23,6 +23,8 @@
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.2.0",
+        "typing-extensions>=4.3.0;python_version<'3.10'",
+        "web-poet @ git+https://git@github.com/scrapinghub/web-poet@facd6d0#egg=web-poet",
```
kmike, line 27: To unblock merging of this PR: what do you think about adding web-poet as a dependency here (e.g. web-poet >= 0.4.0), but specifying the commit in tox.ini only? It seems this particular version is only needed to for tests, and a previous release would work fine, from the user point of view.

zyte_common_items/util.py
kmike: How would it be imported from type_extensions in Python 3.9, if they were only installed for Python < 3.9?

setup.py
BurnzZ: Yep, that should be fine. 👍 https://github.com/zytedata/zyte-common-items/pull/11/commits/17f47b5328fb714911d641963c97943ff571db6d

I think if the user has installed `0.4.0` but then encounters a **mypy** error like `Skipping analyzing "web_poet": module is installed, but missing library stubs or py.typed marker`, then the `web-poet` version could simply be bumped in their project.

zyte_common_items/util.py
BurnzZ: I see 🤔 

Could it be due to GitHub Action's cache? In this commit https://github.com/zytedata/zyte-common-items/commit/3e53d6154ce7e5f1f9d0ec85f7768c100de3ad7f#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R26, it was installed on all versions before `python_version<'3.9'` was incorrectly specified.

BurnzZ: I took a look into this again, trying it out locally while ensuring to delete all caches but `typing-extensions` was still being installed.

Turns out that `pytest-my-testing` is the one that installs it. 😅 https://github.com/davidfritzsche/pytest-mypy-testing/blob/031514ff6ecd5bdf4d11ff238c14d4801b5e47f3/requirements.txt#L169

kmike: Thanks for checking it @BurnzZ!

tests/test_mypy.py
```diff
@@ -0,0 +1,163 @@
...
+)
+
+URL = "https://www.some-page/"
+RESPONSE = HttpResponse(URL, b"")
+
+
+@pytest.mark.mypy_testing
+def test_assignment_image():
+    img_1 = Image(url=RESPONSE.url)
+    img_1.url = RESPONSE.url  # E: Incompatible types in assignment (expression has type "ResponseUrl", variable has type "str")
```
kmike, line 40: Hm, why does it raise a mypy error? Isn't the idea to support conversion on assignment as well?

BurnzZ: I've been thinking about defining `Image` as:

```python
@attr.define
class Image(Item):
    url: Union[str, _Url] = attrs.field(converter=url_to_str)
```

However, this part of the code ([reference](https://github.com/zytedata/zyte-common-items/blob/main/zyte_common_items/base.py#L89)) doesn't allow `Union`s in the field types and would raise a `ValueError`. This constraint is needed to be deterministic when converting using `.from_dict()` and `.from_list()`.

One approach we could take to is to have the `web-poet` URL classes inherit from `str` but this would affect our future plans of using `yarl` or `w3lib` in the future. 

Another alternative is to update our [code ](https://github.com/zytedata/zyte-common-items/blob/main/zyte_common_items/base.py#L89) to accept `Union`s but follows a priority for conversion (e.g. `Union[str, int]` would try to auto-convert the value to `str` first and then `int`). 

Perhaps there are better ways to do this so I'll continue to explore them. Kindly let me know if you also think of one. :) 

Nonetheless, despite this **mypy** error, the auto-conversion of `web-poet`'s URL classes into `str` is still supported as seen in the [conversion tests](https://github.com/zytedata/zyte-common-items/pull/11/files#diff-1de3b8c60439d23d5e65480ab44d0329f42a1312a85497580aaa3fd5e9ac750eR50-R59).

kmike: I was hoping that mypy + attrs are smart enough to look into convertors: url attribute must be str (Union would be wrong, it must be str), but an assignment of _Url should work, because it goes through convertor, which returns str. So, I was expecting type validation to use annotations of url_to_str arguments, not annotation of the url attribute.

If that's not working, shouldn't we mark the test as xfail?

BurnzZ: After looking into this further, it would seem that it's an upstream issue from mypy itself. Marked these specific test cases as xfail and added links to the relevant GH issues in the code comments. https://github.com/zytedata/zyte-common-items/pull/11/commits/635b30daa2106e5448dd8c4ff9a41e9e009960b5

zyte_common_items/util.py
```diff
@@ -54,3 +60,15 @@ def get_origin(tp) -> Tuple:
...
+def is_url_obj(url: Union[str, _Url]) -> TypeGuard[_Url]:
+    return True if isinstance(url, _Url) else False
+
+
+def url_to_str(url: Union[str, _Url]) -> str:
+    if is_url_obj(url):
+        return str(url)
+    elif isinstance(url, str):
+        return url
+    raise ValueError(f"Value of {url} cannot be converted to string.")
```
Gallaecio, line 74: :lipstick: 

```suggestion
    raise ValueError(f"{url!r} is neither a string nor an instance of RequestURL or ResponseURL.")
```

Gallaecio (approved):

zyte_common_items/util.py
```diff
@@ -54,3 +60,15 @@ def get_origin(tp) -> Tuple:
...
+def is_url_obj(url: Union[str, _Url]) -> TypeGuard[_Url]:
+    return True if isinstance(url, _Url) else False
+
+
+def url_to_str(url: Union[str, _Url]) -> str:
+    if is_url_obj(url):
+        return str(url)
+    elif isinstance(url, str):
+        return url
+    raise ValueError(f"{url!r} is neither a string nor an instance of RequestURL or ResponseURL.")
```
kmike, line 74: What do you think about simplifying it to something like this?

```suggestion
def url_to_str(url: Union[str, _Url]) -> str:
    if not isinstance(url, (str, _Url)):
        raise ValueError(f"{url!r} is neither a string nor an instance of RequestURL or ResponseURL.")
    return str(url)
```

`is_url_obj` doesn't look required; `_Url` is visible in url_to_str signature, so I'm not sure an extra abstraction helps. 

Alternatively, it can also be simplified by dropping is_url_obj, and checking `isinstance(url, _Url)` directly.

Even if util.py is considered an internal module, it's not underscored, and is_url_obj is not underscored, so it becomes a part of public API; it doesn't seem required.

kmike (approved):
👍

zyte_common_items/util.py
BurnzZ: Good idea. The `TypeGuard` approach was an attempt to help mypy discern the return type of the converter via type narrowing. But it's not being used right now because of the upstream issues between mypy and attr's converters.

PR #19, opened 2022-11-01
Add ProductPage and ProductListPage

Gallaecio:
Note: includes unrelated https://github.com/zytedata/zyte-common-items/pull/19/commits/34528b2d720f83ad745a69068e75f65f47ce9374 and https://github.com/zytedata/zyte-common-items/pull/19/commits/2d1cd89d1dc7ba16524df311a5d2391989f84bf3

To do:
- [x] Wait for a web-poet release that includes https://github.com/scrapinghub/web-poet/pull/109, then update this pull request to rely on that web-poet version rather than a Git branch.


zyte_common_items/page_objects.py
```diff
@@ -0,0 +1,31 @@
...
+from .items import Product, ProductList
+
+
+@attrs.define
+class Page(ItemPage):
+    _url: ResponseUrl
+
+    @field
+    def metadata(self) -> Metadata:
+        """…"""
```
BurnzZ, line 16:
```suggestion
        """Marks date and time when the page object created the item."""
```

tests/test_page_objects.py
```diff
@@ -0,0 +1,64 @@
...
+    url = ResponseUrl("https://example.com/books/1")
+    html = b"""
+    <!DOCTYPE html>
+    <html>
+        <body>
+            <h1>Foo</h1>
+        </body>
+    </html>
+    """
+    response = HttpResponse(url=url, body=html)
```
BurnzZ, line 53:
```suggestion
    response = HttpResponse(_url=url, body=html)
```

tests/test_page_objects.py
```diff
@@ -0,0 +1,64 @@
...
+    assert datetime_before <= page_datetime <= datetime_after
+
+
+@pytest.mark.asyncio
+async def test_example():
+    datetime_before = datetime.utcnow().replace(microsecond=0)
+
+    @attrs.define
+    class BookPage(ProductPage):
+        response: HttpResponse
```
BurnzZ, line 38: It occurred to me that majority of the user's use case would always need the `HttpResponse` dependency.

I'm thinking if we should have `ProductPage` inherit from `web_poet.WebPage` instead of `web_poet.ItemPage` to already include `HttpResponse` from the get go. This could save the user one line of code like this one.

An issue with `web_poet.WebPage` though is that it doesn't mark such dependency to be `Optional[HttpResponse]`. So in cases where users don't need the response, *mypy* would complain if they pass `None`. However, they could simply pass a `HttpResponse` with some fake values to quiet *mypy* down. In any case, I think this would be an edge case since I expect users to be always using the response via `self.css()` and `self.xpath()` inside the PO.

Thoughts @Gallaecio @kmike ?

tests/test_page_objects.py
BurnzZ: Ahh I marked the wrong line.

tests/test_page_objects.py
```diff
@@ -0,0 +1,64 @@
...
+    <!DOCTYPE html>
+    <html>
+        <body>
+            <h1>Foo</h1>
+        </body>
+    </html>
+    """
+    response = HttpResponse(url=url, body=html)
+
+    item = await BookPage(url=url, response=response).to_item()
```
BurnzZ, line 55:
```suggestion
    item = await BookPage(_url=url, response=response).to_item()
```

This is tricky since the field name should be `url` so the dependency should not have the same name with it. But at the same time, passing `_url` looks weird. 🤔

tests/test_page_objects.py
Gallaecio: If a page object uses browser HTML, that could be a problem, as you cannot remove attrs attributes, i.e. both response body and browser HTML would be provided through dependency injection.

zyte_common_items/page_objects.py
Gallaecio: This is a leftover from testing https://github.com/scrapinghub/web-poet/issues/99, I need to remove it altogether, at the moment the docs need to be written on the reStructuredText files instead.

Gallaecio:
Today I Learned attrs expects underscore attributes to be passed to `__init__` without the underscore.

wRAR (approved):

README.rst
```diff
@@ -20,12 +20,12 @@ zyte-common-items
 
 .. description starts
 
-``zyte-common-items`` is a Python 3.7+ library of item classes used by Zyte_ to
-normalize different types of data extracted from websites.
-
-It can be used in custom data extraction code for normalization purposes,
-maximizing opportunities for code reuse.
+``zyte-common-items`` is a Python 3.7+ library of item_ and `page object`_
```
kmike, line 23: a side note: we might need to think about a better name for this package

Gallaecio: zyte-poet?

docs/reference/page-objects.rst
```diff
@@ -0,0 +1,43 @@
...
+
+   .. data:: metadata
+      :type: zyte_common_items.Metadata
+
+      Data extraction process metadata.
+
+      :attr:`~zyte_common_items.Metadata.dateDownloaded` is set to the current
+      UTC date and time.
+
+      :attr:`~zyte_common_items.Metadata.probability` is not set.
```
kmike, line 38: I think we should set it to 1.0

tests/test_page_objects.py
kmike: I think both points are valid. It seems we need to have both classes: with and without HttpResponse defined as a dependency. The main question is where we should have them. 

We should definitely provide a class without HttpResponse dependency in this repo. 

For a ProductPage with HttpResponse dependency there are 2 options:

1) Add another class here. Probably rename ProductPage to BaseProductPage, and add ProductPage with response dependency.
2) Encourage users to subclass ProductPage defined here, not use it directly.

Gallaecio: Isn’t 2 the current scenario? Don’t users need to subclass ProductPage either way to define additional fields?

BurnzZ: Good point about having `browserHtml` as it could possibly be the 2nd most common use case.

Option 1 with `BaseProductPage` and `ProductPage` sounds good to me. 👍

README.rst
BurnzZ: What about `zyte-commons`? or `zyte-reuse`? `zyte-building-blocks`?

Something that could possibly accommodate other things in the future that we haven't thought about.

Gallaecio: Like those first 2.

Playing with `zyte-reuse`, I did thought of `rezyte`, which would be very much in line with anything “poet”, but I still like your suggestions much better :smile:

tests/test_page_objects.py
kmike: > Isn’t 2 the current scenario?

I mean that we might need to advice users to create their own base ProductPage, which is a subclass of zyte_common_items.ProductPage, with the response: HttpResponse defined.

zyte_common_items/__init__.py
```diff
@@ -12,3 +12,4 @@
     Metadata,
 )
 from .items import Product, ProductFromList, ProductList, ProductVariant
+from .page_objects import Page, ProductListPage, ProductPage
```
kmike, line 15: What do you think about using "pages" convention for modules and packages with page objects?

tests/test_page_objects.py
Gallaecio: [Went for option 1](https://github.com/zytedata/zyte-common-items/pull/19/commits/a14a9bdc809c5263b6ff95e70696c3e20dbba3ed).

docs/usage/pages.rst
```diff
@@ -0,0 +1,38 @@
...
+:attr:`~zyte_common_items.Product.metadata`, a
+:attr:`~zyte_common_items.Product.name`, and a
+:attr:`~zyte_common_items.Product.url`:
+
+.. code-block:: python
+
+   import attrs
+   from zyte_common_items import ProductPage
+
+    @attrs.define
```
BurnzZ, line 33:
```suggestion
```

BurnzZ (approved):

zyte_common_items/pages.py
```diff
@@ -0,0 +1,53 @@
...
+import attrs
+from web_poet import ItemPage, ResponseUrl, Returns, WebPage, field
+
+from .components import Metadata
+from .items import Product, ProductList
+
+
+@attrs.define
+class BasePage(ItemPage):
+    _url: ResponseUrl
```
Gallaecio, line 12: Imagine you write a page object class that needs a _request_ URL, e.g. extracts some product ID from the request URL, and uses an additional request to fetch the product data from a JSON API.

For that scenario, subclasssing `BasePage` directly or indirectly will cause your page object class to get an extra, unnecessary `ResponseUrl` input, which requires an actual request (or more) to resolve any redirect from `RequestUrl`.

While I don’t think this should be a blocker here, we might want to try and think how we want to address this eventually.

For example, we could eventually have some method in `ItemPage` that allows querying for an input that may not exist, and then we can remove `_url` from here, so that `ResponseUrl` no longer becomes a required input, and redefine the url field in a way that uses any `ResponseUrl` if defined, or raises `NotImplementedError` otherwise:

```python
    @field
    def url(self) -> str:
        try:
            url = self.get_input(ResponseUrl)
        except ValueError:
            raise NotImplementedError
        return str(url)
```

Note: I assume `self.get_input(ResponseUrl)` would work if e.g. `HttpResponse` exists as input, since `ResponseUrl` can be built from it.

A page object class without a `ResponseUrl`-compatible input would have to reimplement the field to return the right URL, or else get `NotImplementedError` raised when `to_item()` is called.

tox.ini
```diff
@@ -1,11 +1,14 @@
...
 
 [testenv]
 deps =
     pytest
+    pytest-asyncio
     pytest-cov
-    pytest-mypy-testing==0.0.11
+    # https://github.com/davidfritzsche/pytest-mypy-testing/issues/35
+    git+https://github.com/davidfritzsche/pytest-mypy-testing.git@031514ff6ecd5bdf4d11ff238c14d4801b5e47f3
+    mypy==0.971
```
wRAR, line 11: Is mypy needed here now?

Gallaecio: I would get installed by the dependency above, but I think I added it here to freeze its version, in line with what we do in Scrapy.

zyte_common_items/pages.py
kmike: I think that could be a big issue. 

It seems there are 2 sides of this issue. 

First is related to the problem at hand. It seems that using RequestUrl to set Product.url may be not precise enough, as product.url should contain response url. But even if it's not precise, I'm not sure the alternative is better (ask users to provide real url which was used when they make a request using HttpClient). Take an example of AJAX request. You don't want URL of the API endpoint in the "url" field: it is weird for the end users, leaking internal details, and also it means that product can't be extracted by the page object if product.url is passed to it as an input, which is against the field definition. So, I think that RequestUrl is a relatively safe choice. 

We might even consider using it as a default for everything. It may be a better alternative than a trap with ResponseUrl triggering download.

Second is a more general issue, which affects more than RequestUrl vs ResponseUrl. There are other cases when there is something which is expensive to get, and something which is easy to get, and you want an expensive version only if it's "free", i.e. you need to do an expensive operation anyways.

Examples:

* ResponseUrl vs RequestUrl
* ResponseUrl from HttpResponse vs url from BrowserHtml (not sure about that, we don't have a separate url for browser now)
* sometimes code is written to work both with httpResponseBody and with browserHtml, and you want to get something which is already available

In andi there is support for [Unions](https://github.com/scrapinghub/andi#union) and [Optional](https://github.com/scrapinghub/andi#optional), so it's possible to write

```py
any_html: BrowserHtml | HttpResponse
response_url: Optional[ResponseUrl]
```

The semantics is "use BrowserHtml if available; if not, use HttpResponse". The issue is with a definition of "available" - it's defined as "can be provided", not as "it is already provided elsewhere, and can be reused with no extra work". So, it doesn't solve the problem. We might want to implement something like

```py
any_html: Reuse[BrowserHtml] | HttpResponse
response_url: Reuse[ResponseUrl] | None
```

It seems this would allow to express what we want here cleanly: base class can have a definition like this:

```py
request_or_response_url: Reuse[ResponseUrl] | RequestUrl
```

So, that'd be my proposal:

1. For the current PR just use RequestUrl everywhere.
2. Think about adding Reuse support, to andi / web-poet / scrapy-poet.

What do you think?

---

There is one issue with RequestUrl, which is related to testing (cc @wRAR). Let's say we have a page object which sets url using self.response.url. We generate test fixtures for it using scrapy-poet. Then, we refactor it to use a base class defined here, and want to run tests to see if it works as before. But it seems it won't run, because dependencies are different - there is no RequestUrl dependency in the fixture. I'm not sure, but it might also apply for ResponseUrl vs HttpResponse. Users may need to do some manual steps in this case, or maybe we can provide some helper script, not sure. 

We may also change the savefixture command to always store RequestUrl dependency, even if it's not required by the page object.

---

What do you think @Gallaecio @wRAR @BurnzZ ?

BurnzZ: I'm not quite sure on using `RequestUrl` everywhere here since it breaks the definition of **url** for Zyte Data Schemas:

> The URL of the final response, after any redirects.

Though I fully agree with the semantics of:

```
request_or_response_url: Reuse[ResponseUrl] | RequestUrl
```

since it's inline with the description of **url** in our schema:

> In case there is no product data on the page or the page was not reached, the returned "empty" item would still contain `url` field and `metadata` field with `dateDownloaded`.

With this, I think we can remove the `_url: ResponseUrl` dependency from `BasePage` and instead have `_BaseMixin` use something like:

```python
import inspect

class _BaseMixin:

    @field
    def url(self) -> str:
        urls = {}

        for name, dep in inspect.signature(self.__init__).parameters.items():
            if dep.annotation is ResponseUrl:
                urls["response_url"] = getattr(self, name)
            elif dep.annotation is HttpResponse:
                urls["response_url"] = getattr(self, name).url
            elif dep.annotation is RequestUrl:
                urls["request_url"] = getattr(self, name)

        url = urls.get("response_url") or urls.get("request_url")
        return str(url)
```

But I'm not sure if introspection is the way to go.

Gallaecio: > For the current PR just use RequestUrl everywhere.

I’m OK with this given we will be working on addressing this with the `Reuse` approach.

But I do wonder if we should try to get the initial implementation to be closer to idea, with something in line with what @BurnzZ suggests above.

kmike: What do you think about th following?

1. Have a base class which depends on RequestUrl, and implements url field using self.request_url
2. Page class (which is a subclass of WebPage) re-implements url field using self.response.url

So, there is no ResponseUrl dependency anywhere, and we don't need to do any inspection. There are 2 downsides:

1. We always use RequestUrl dependency, even if it's not needed. I might actually not be a bad thing, as it may testing tool to work better in case of refactoring, and it should be free to get RequestUrl.
2. If a class uses BrowserHtml, it'd need to re-implement url method again to use ResponseUrl. I think that's also fine. We don't have a provider for BrowserHtml yet, and we'd need to do some work to figure out how ResponseUrl is going to work in this case.

Gallaecio: https://github.com/zytedata/zyte-common-items/pull/19/commits/965b433d61f45065477d731af88299c873118efc

zyte_common_items/pages.py
```diff
@@ -0,0 +1,53 @@
...
+    @field
+    def metadata(self) -> Metadata:
+        return Metadata(
+            dateDownloaded=f"{datetime.utcnow().isoformat(timespec='seconds')}Z",
+            probability=1.0,
+        )
+
+    @field
+    def url(self) -> str:
+        return str(self._get_response_url())
```
kmike, line 23: What's the purpose of this indirection? The alternative is to add url field to BasePage and Page classes.

kmike (approved):
Thanks @Gallaecio!

I had small reservations about Page not being a subclass of BasePage, but it seems it's better this way, and we can change it if that's not the case.

PR #21, opened 2023-01-24
POC: Using `web_poet.Unset`

BurnzZ:
Uses: https://github.com/scrapinghub/web-poet/pull/116/

The tests for adapters are failing since `Unset` is not treated like `None` which results in their values not being dropped in the resulting dict.


Gallaecio:
> The tests for adapters are failing since `Unset` is not treated like `None` which results in their values not being dropped in the resulting dict.

[It should be easy to address](https://github.com/zytedata/zyte-common-items/blob/main/zyte_common_items/adapter.py#L12).

PR #25, opened 2023-02-17
Improve from_dict/from_list error reporting

Gallaecio:


zyte_common_items/base.py
```diff
@@ -102,19 +102,30 @@ def _apply_field_types_to_sub_fields(cls, item: Dict):
...
+                    )
+                item[key] = cls.from_dict(value)
+            for key, cls in (from_list or {}).items():
+                value = item.get(key)
+                if value is not None and not isinstance(value, list):
+                    path = f"{cls.__module__}.{cls.__qualname__}"
+                    raise ValueError(
+                        f"Expected {key!r} to be a list of dicts with fields "
+                        f"from {path}, got {value!r}."
+                    )
```
BurnzZ, line 128: I think there's a potential to put this `if`\-statement and the raised error into a single function since these 2 instances for both dict and list are quite the same.

BurnzZ (commented):
This makes debugging so much easier! 💯

kmike (approved):
👍

zyte_common_items/base.py
Gallaecio: We could do:

```python
def _fill_from_nested_fields(
    *,
    item: dict,
    fields: dict,
    load_method: str,
    expected_type: Union[Dict, List],
    expected_type_name: str,
):
    for key, cls in (fields or {}).items():
        value = item.get(key)
        if value is not None and not isinstance(value, expected_type):
            path = f"{cls.__module__}.{cls.__qualname__}"
            raise ValueError(
                f"Expected {key!r} to be a {expected_type_name} with fields "
                f"from {path}, got {value!r}."
            )
        item[key] = getattr(cls, load_method)(value)

# …


            _fill_from_nested_fields(
                item=item,
                fields=from_dict,
                load_method="from_dict",
                expected_type=dict,
                expected_type_name="dict",
            )
            _fill_from_nested_fields(
                item=item,
                fields=from_list,
                load_method="from_list",
                expected_type=list,
                expected_type_name="list of dicts",
            )
```

But I am unsure whether or not it is better.

BurnzZ (approved):

PR #27, opened 2023-03-16
Business place objects

rromaniuc:
related to https://zyte.atlassian.net/browse/MAJ-354


tests/test_components.py
```diff
@@ -21,3 +32,45 @@ def test_metadata_default_values():
...
+def test_anchor_optional_fields():
+    Link(text="foo")
+    Link(url="https://example.com")
+
+
+def test_business_place_metadata_default_values():
+    metadata = BusinessPlaceMetadata()
+    assert metadata.dateDownloaded is None
+    assert metadata.probability is None
+    assert metadata.searchText is None
```
Gallaecio, line 46: @kmike I did not expect the metadata field to be different in different schemas. How do you want us to handle this on the code side? Is this approach of a item-type-specific metadata class OK? Should we try to find a generic name for this metadata variant that includes search terms?

tests/test_components.py
```diff
@@ -21,3 +32,45 @@ def test_metadata_default_values():
...
+
+
+def test_address():
+    Address(addressRaw="2440 Hoonani Rd, Koloa, HI 96756, US")
+    Address(streetAddress="2440 Hoonani Rd, Koloa, HI 96756")
+    Address(addressCity="Koloa")
+    Address(addressLocality="Poipu")
+    Address(addressRegion="Kauaʻi County")
+    Address(addressCountry="US")
+    Address(postalCode="HI 96756")
```
Gallaecio, line 56:
```suggestion
    Address(postalCode="96756")
```

We also seem to be missing `postalCodeAux="HI"`.

tests/test_components.py
```diff
@@ -21,3 +32,45 @@ def test_metadata_default_values():
...
+    StarRating(raw="4-star hotel")
+    StarRating(ratingValue=4)
+
+
+def test_parent_place():
+    ParentPlace(name="Gateway Plaza", placeId="ChIJCVDfcEKuEmsR7_zYc0_GBtQ")
+
+
+def test_opening_hours():
+    OpeningHours(dayOfWeek="Sunday", opens="9:00", closes="12:00")
```
Gallaecio, line 76: I wonder if we should find a name that does not match the array property where this would be but 1 item. `OpeningHoursItem`? `OpeningHoursPeriod`? `OpeningPeriod`?

tests/test_items.py
```diff
@@ -107,6 +115,51 @@
     "products": [ProductFromList()],
 }
 
+_BUSINESS_PLACE_MIN_KWARGS: dict = {
+    "url": "https://www.google.com/maps?cid=8273826543880430215"
```
Gallaecio, line 119:
```suggestion
    "url": "https://place-listing.example/place/12345"
```

tests/test_items.py
```diff
@@ -107,6 +115,51 @@
...
+    ],
+    "additionalProperties": [
+        AdditionalProperty(name="Popular for", value="Lunch,Dinner")
+    ],
+    "address": Address(addressRaw="2440 Hoonani Rd, Koloa, HI 96756, US"),
+    "reservationAction": Anchor(name="Reserve a table", url="https://example.com"),
+    "categories": ["Hotel", "Wedding venue"],
+    "description": "Oceanfront resort offering Hawaiian dining, plus 2 outdoor pools & a fitness center.",
+    "features": ["Pool", "Free Parking"],
+    "map": "https://www.google.com/maps?cid=8273826543880430215",
```
Gallaecio, line 138:
```suggestion
    "map": "https://map.example/place/12345",
```

zyte_common_items/components.py
```diff
@@ -127,6 +128,25 @@ class Link(Item):
...
 
 
+@attrs.define(kw_only=True)
+class Anchor(Item):
+    """A link from a webpage to another webpage.
+
+    Only difference between :class:`Link <zyte_common_items.components.Link> and this class
+    is the attributes "name" and "text".
+
+    Can be merged in the future but require changing the standard schemas.
```
Gallaecio, line 138: This note should not be in code, I think.

zyte_common_items/components.py
```diff
@@ -127,6 +128,25 @@ class Link(Item):
     )
 
 
+@attrs.define(kw_only=True)
+class Anchor(Item):
```
Gallaecio, line 132: Anchor makes me think of the targets of URL fragments (#foo).

Maybe something like `NamedLink` would work?

zyte_common_items/util.py
```diff
@@ -65,3 +66,15 @@ def url_to_str(url: Union[str, _Url]) -> str:
...
+
+
+def validate_timezone(value: Optional[str] = None):
+    """Validates the value against valid ZoneInfo or optional"""
+    if not value:
+        return None
+    try:
+        _ = pytz.timezone(value)
+    except pytz.UnknownTimeZoneError:
+        raise ValueError(f"{value} is not a valid timezone")
```
Gallaecio, line 78: Do we know for a fact that pytz.timezone only allows tzdata values?

rromaniuc: Did some research, only thing better than pytz is ZoneInfo but requires python3.9+. 
We could switch to ZoneInfo and use backports.zoneinfo for python3.7 / 3.8.

PR #30, opened 2023-03-17
Fix BaseMixin field leak

Gallaecio:
Closes #29.

https://github.com/scrapinghub/web-poet/pull/153 is related, but does not actually affect zyte-common-items since when we inherit from `_BaseMixin` it is the first base class and the rest (ItemPage, WebPage) do not define fields of their own.


Nykakin (approved):
Tested and works for me.

PR #37, opened 2023-05-02
ProductNavigation Item

VMRuiz:
Hello, 

This is my first contribution to this repo, so please make sure I didn't miss any detail. 

Also, I named Navigation to the object containing all details of the HTTP requests, but maybe it should be named `HTTPNavigation` or `HTTPRequest` ? That would open the possibility of adding other types of request in the future based on Browser rendering like `BrowserNavigation` or `BrowserRequest` ?

### DIFF
--- a/tests/test_components.py
+++ b/tests/test_components.py
@@ -11,6 +11,8 @@ from zyte_common_items import (
     ParentPlace,
     RealEstateArea,
     StarRating,
+    Header,
+    Navigation,
 )
 
 
@@ -83,4 +85,24 @@ def test_real_estate_area():
     RealEstateArea(value=123.45, unitCode="SQMT", raw="123.45 square meters")
     RealEstateArea(
         value=123.45, unitCode="SQMT", raw="123.45 square meters", areaType="LOT"
+    )
+
+
+def test_headers():
+    Header(name="Connection", value="keep-alive")
+    Header(name="Content-Type", value="text/html; charset=utf-8")
+
+
+def test_navigation():
+    Navigation(url="https://example.com")
+    headers = [
+        Header(name="Content-Type", value="application/x-www-form-urlencoded"),
+        Header(name="Host", value="foo.example"),
+    ]
+    Navigation(
+        name="Post Test",
+        url="https://example.com/test",
+        method="POST",
+        body="field1=value1&field2=value2",
+        headers=headers,
     )
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -25,6 +25,9 @@ from zyte_common_items import (
     RealEstate,
     RealEstateArea,
     StarRating,
+    ProductNavigation,
+    Navigation,
+    Header,
 )
 
 _PRODUCT_FROM_LIST_ALL_KWARGS: dict = {
@@ -211,6 +214,62 @@ _REAL_ESTATE_ALL_KWARGS: dict = {
     ),
 }
 
+_PRODUCT_NAVIGATION_MIN_KWARGS: dict = {"url": "https://example.com/real-estate/12345"}
+
+_PRODUCT_NAVIGATION_ALL_KWARGS: dict = {
+    **_PRODUCT_NAVIGATION_MIN_KWARGS,
+    "categoryName": "Swiss Watches",
+    "subCategories": [
+        Navigation(
+            **{
+                "url": "http://books.toscrape.com/catalogue/category/books/",
+                "method": "POST",
+                "body": "YmFzZTY0LWVuY29kZWQ=",
+                "headers": [Header(name="content-type", value="text/json")],
+                "name": "Travel",
+                "metadata": Metadata(probability=0.99),
+            }
+        ),
+        Navigation(
+            **{
+                "url": "http://books.toscrape.com/catalogue/category/books/mystery_3/index.html",
+                "name": "Mystery",
+                "metadata": Metadata(probability=0.97),
+            }
+        ),
+    ],
+    "items": [
+        Navigation(
+            **{
+                "url": "http://books.toscrape.com/catalogue/in-her-wake_980",
+                "method": "POST",
+                "body": "YmFzZTY0LWVuY29kZWQ=",
+                "headers": [Header(name="content-type", value="text/json")],
+                "name": "In Her Wake",
+                "metadata": Metadata(probability=0.99),
+            }
+        ),
+        Navigation(
+            **{
+                "url": "http://books.toscrape.com/catalogue/how-music-works_979/index.html",
+                "name": "How Music Works",
+                "metadata": Metadata(probability=0.98),
+            }
+        ),
+    ],
+    "nextPage": Navigation(
+        **{
+            "url": "http://books.toscrape.com/catalogue/page-3.html",
+            "name": "Next page",
+            "method": "POST",
+            "body": "Im9rIg==",
+            "headers": [Header(name="content-type", value="text/json")],
+        }
+    ),
+    "pageNumber": 5,
+    "metadata": Metadata(dateDownloaded="2022-12-31T13:01:54Z"),
+}
+
 
 def test_product_all_fields():
     product = Product(**_PRODUCT_ALL_KWARGS)
@@ -322,3 +381,27 @@ def test_real_estate_missing_fields():
         del incomplete_kwargs[required_field]
         with pytest.raises(TypeError):
             RealEstate(**incomplete_kwargs)
+
+
+def test_product_navigation_all_fields():
+    product_navigation = ProductNavigation(**_PRODUCT_NAVIGATION_ALL_KWARGS)
+    for field in list(_PRODUCT_NAVIGATION_ALL_KWARGS):
+        assert (
+            getattr(product_navigation, field) == _PRODUCT_NAVIGATION_ALL_KWARGS[field]
+        )
+
+
+def test_product_navigation_min_fields():
+    product_navigation = ProductNavigation(**_PRODUCT_NAVIGATION_MIN_KWARGS)
+    for field in list(_PRODUCT_NAVIGATION_ALL_KWARGS):
+        if field in _PRODUCT_NAVIGATION_MIN_KWARGS:
+            continue
+        assert getattr(product_navigation, field) is None
+
+
+def test_product_navigation_missing_fields():
+    for required_field in list(_PRODUCT_NAVIGATION_MIN_KWARGS):
+        incomplete_kwargs: dict = copy(_PRODUCT_NAVIGATION_MIN_KWARGS)
+        del incomplete_kwargs[required_field]
+        with pytest.raises(TypeError):
+            ProductNavigation(**incomplete_kwargs)
--- a/zyte_common_items/__init__.py
+++ b/zyte_common_items/__init__.py
@@ -18,6 +18,8 @@ from .components import (
     ParentPlace,
     RealEstateArea,
     StarRating,
+    Header,
+    Navigation,
 )
 from .items import (
     BusinessPlace,
@@ -26,6 +28,7 @@ from .items import (
     ProductList,
     ProductVariant,
     RealEstate,
+    ProductNavigation,
 )
 from .pages import (
     BaseBusinessPlacePage,
@@ -38,4 +41,5 @@ from .pages import (
     ProductListPage,
     ProductPage,
     RealEstatePage,
+    ProductNavigationPage
 )
--- a/zyte_common_items/components.py
+++ b/zyte_common_items/components.py
@@ -1,5 +1,5 @@
 """Classes for data nested within items."""
-from typing import Optional
+from typing import Optional, List, NewType
 
 import attrs
 
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
 
     #: Area in the raw format, as it appears on the website.
     raw: str
+
+
+@attrs.define(kw_only=True)
+class Header(Item):
+    """A http header value"""
+
+    #: Name of header
+    name: str
+
+    #: Value of the header
+    value: str
+
+
+@attrs.define(kw_only=True)
+class Navigation(Item):
+    """Describe a HTTP request to load a page"""
+
+    #: Name of the category/page to which the navigation takes
+    name: Optional[str] = None
+
+    #: HTTP URL
+    url: str = attrs.field(converter=url_to_str)
+
+    #: HTTP Method
+    #:
+    #: Values: "GET" "POST" "PUT" "DELETE" "OPTIONS" "TRACE" "PATCH"
+    method: str = "GET"
+
+    #: HTTP Request Body
+    #:
+    #: Format:
+    #: - Base64 Encoded
+    body: Optional[str] = None
+
+    #: HTTP Headers
+    headers: Optional[List[Header]] = None
+
+    #: Data extraction process metadata.
+    metadata: Optional[Metadata] = None
--- a/zyte_common_items/items.py
+++ b/zyte_common_items/items.py
@@ -20,6 +20,7 @@ from zyte_common_items.components import (
     ParentPlace,
     RealEstateArea,
     StarRating,
+    Navigation
 )
 from zyte_common_items.util import url_to_str
 
@@ -711,4 +712,38 @@ class RealEstate(Item):
     virtualTourUrl: Optional[str] = None
 
     #: Contains metadata about the data extraction process.
+    metadata: Optional[Metadata] = None
+
+@attrs.define(kw_only=True)
+class ProductNavigation(Item):
+    """Represents navigations for a page of product listings on a e-commerce website """
+
+    #: Main URL from which the data has been extracted.
+    url: str = attrs.field(converter=url_to_str)
+
+    #: Name of the category/page with the product list.
+    #:
+    #: Format:
+    #:
+    #: - trimmed (no whitespace at the beginning or the end of the description string),
+    categoryName: Optional[str] = None
+
+    #: List of sub-catetegory links found on the page. The order of the links reflects their position on the page.
+    subCategories: Optional[List[Navigation]] = None
+
+    #: List of product links found on the page which belond to the page category. The order of the links reflects their position on the page.
+    items: Optional[List[Navigation]] = None
+
+    #: The navigation to the next page, if available.
+    nextPage: Optional[Navigation] = None
+
+    #: Number of the current page.
+    #:
+    #: It should only be extracted if the webpage shows a page number.
+    #:
+    #: It must be 1-based. For example, if the first page of a listing is
+    #: numbered as 0 on the website, it should be extracted as `1` nonetheless.
+    pageNumber: Optional[int] = None
+
+    #: Data extraction process metadata.
     metadata: Optional[Metadata] = None
--- a/zyte_common_items/pages.py
+++ b/zyte_common_items/pages.py
@@ -5,7 +5,7 @@ from web_poet import ItemPage, RequestUrl, Returns, WebPage, field
 from web_poet.fields import FieldsMixin
 
 from .components import Metadata
-from .items import BusinessPlace, Product, ProductList, RealEstate
+from .items import BusinessPlace, Product, ProductList, RealEstate, ProductNavigation
 
 
 class _BaseMixin(FieldsMixin):
@@ -42,6 +42,10 @@ class BaseRealEstatePage(BasePage, Returns[RealEstate]):
     pass
 
 
+class BaseProductNavigationPage(BasePage, Returns[ProductNavigation]):
+    pass
+
+
 @attrs.define
 class Page(_BaseMixin, WebPage):
     @field
@@ -62,4 +66,8 @@ class BusinessPlacePage(Page, Returns[BusinessPlace]):
 
 
 class RealEstatePage(Page, Returns[RealEstate]):
+    pass
+
+
+class ProductNavigationPage(Page, Returns[ProductNavigation]):
     pass


tests/test_components.py
```diff
@@ -84,3 +86,23 @@ def test_real_estate_area():
...
     )
+
+
+def test_headers():
+    Header(name="Connection", value="keep-alive")
+    Header(name="Content-Type", value="text/html; charset=utf-8")
+
+
+def test_navigation():
+    Navigation(url="https://example.com")
```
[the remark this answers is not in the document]
VMRuiz: I guess that would depend if we will implement Browser Request on the future and how we do it? 
Will it be part of the current `Navigation` object? Will it be an independent class?

If we use a single class for all then `Request` looks good. Otherwise, we should have some references to http and browser on the class name.

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
 
     #: Contains metadata about the data extraction process.
     metadata: Optional[Metadata] = None
+
+
+@attrs.define(kw_only=True)
+class ProductNavigation(Item):
+    """Represents navigations for a page of product listings on a e-commerce website"""
+
+    #: Main URL from which the data has been extracted.
```
BurnzZ, line 722:
```suggestion
    #: Main URL from which the data is extracted.
```

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    url: str = attrs.field(converter=url_to_str)
+
+    #: Name of the category/page with the product list.
+    #:
+    #: Format:
+    #:
+    #: - trimmed (no whitespace at the beginning or the end of the description string),
+    categoryName: Optional[str] = None
+
+    #: List of sub-catetegory links found on the page. The order of the links reflects their position on the page.
```
BurnzZ, line 732:
```suggestion
    #: List of sub-category links ordered by their position in the page.
```

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    subCategories: Optional[List[Navigation]] = None
+
+    #: List of product links found on the page which belond to the page category. The order of the links reflects their position on the page.
```
BurnzZ, line 735:
```suggestion
    #: List of product links found on the page category ordered by their position in the page.
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+
+    #: Name of the category/page to which the navigation takes
+    name: Optional[str] = None
+
+    #: HTTP URL
+    url: str = attrs.field(converter=url_to_str)
+
+    #: HTTP Method
+    #:
+    #: Values: "GET" "POST" "PUT" "DELETE" "OPTIONS" "TRACE" "PATCH"
```
BurnzZ, line 313: For discussion: Should this be strictly in correspondence with what Zyte API supports? or should it be also be open to other HTTP Methods even when Zyte API doesn't support it (e.g. HEAD)

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+
+    #: Value of the header
+    value: str
+
+
+@attrs.define(kw_only=True)
+class Navigation(Item):
+    """Describe a HTTP request to load a page"""
+
+    #: Name of the category/page to which the navigation takes
```
BurnzZ, line 305:
```suggestion
    #: Name of the page which describes the URL.
```

tests/test_components.py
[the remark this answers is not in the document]
BurnzZ: I think we can discern them via `HttpRequest` and `BrowserRequest`. Probably something like https://github.com/scrapy/scrapy/blob/master/scrapy/http/response/html.py#L11

zyte_common_items/components.py
Gallaecio: [HEAD is coming to Zyte API soon](https://github.com/zytedata/zde-api-server/pull/341), so it may be worth adding.

Not sure if `CONNECTION` will ever make sense here. Not that I think other methods here make sense for navigation, but I guess including them is safer.

VMRuiz: I would add all methods defined by the HTTP protocol. Even if Zyte API, Scrapy, or whatever underlying tech don't support it, the items definitions should be  independent from the underlying tech stack.

tests/test_components.py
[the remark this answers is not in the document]
kmike: Let's clarify what we mean by "Browser Request" :) I think there are at least 2 ways to understand it:

1. When you return a browser request e.g. `BrowserRequest(url=...)`, then browser should be used to render the *target* page.
2. There are some additional request parameters, like `actions` or `js_code`, which you need to execute on the *original page* to get to the next page. For example, click on some button.

What of these use cases do you have in mind @VMRuiz @Gallaecio @BurnzZ ?

My 2c: I'm not sure (1) is needed. It seems it should be a decision made separately if we need to use a browser on the *target* page or not, it's not something you should decide during the navigation. Also, it doesn't solve a case when there is no real link available on a page, but there is a way to go e.g. to a next page by performing some JS action.

(2) may require some thought. It should work, but it'd mean that we might be rendering the page 1 more time than necessary, though this can be ok.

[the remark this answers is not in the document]
kmike: If we're talking about (2), then both Request and Navigation looks ok to me, with the idea that we can extend them in future, adding a field like `browserActions`.

tests/test_components.py
```diff
@@ -84,3 +86,23 @@ def test_real_estate_area():
...
+    headers = [
+        Header(name="Content-Type", value="application/x-www-form-urlencoded"),
+        Header(name="Host", value="foo.example"),
+    ]
+    Navigation(
+        name="Post Test",
+        url="https://example.com/test",
+        method="POST",
+        body="field1=value1&field2=value2",
```
kmike, line 106: Isn't the body base64-encoded in the schema?

tests/test_components.py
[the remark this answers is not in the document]
kmike: We also have the existing Link data type, which is similar; that's another name we can use.

tests/test_components.py
kmike: We don't need to address it in this PR, but it'd be nice to have a more user-friendly way to set the body in Navigation items. E.g. auto-convert bytes to base64. But it's not 100% clear if we should use attrs convertors or web-poet processors for this, so to speed up the merge we can address it separately.

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    #: Number of the current page.
+    #:
+    #: It should only be extracted if the webpage shows a page number.
+    #:
+    #: It must be 1-based. For example, if the first page of a listing is
+    #: numbered as 0 on the website, it should be extracted as `1` nonetheless.
+    pageNumber: Optional[int] = None
+
+    #: Data extraction process metadata.
+    metadata: Optional[Metadata] = None
```
kmike, line 750: I think we need to clarify the definition of metadata fields. E.g. in the schema @lopuhin published there is no probability in the metadata, so it's not the same data structure as e.g. for Product. The metadata type for subCategories / items might be different as well. 

TBH, I'd prefer to use less distinct Metadata classes.

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    categoryName: Optional[str] = None
+
+    #: List of sub-category links ordered by their position in the page.
+    subCategories: Optional[List[Navigation]] = None
+
+    #: List of product links found on the page category ordered by their position in the page.
+    items: Optional[List[Navigation]] = None
+
+    #: The navigation to the next page, if available.
+    nextPage: Optional[Navigation] = None
```
kmike, line 739: If I'm reading the schema correctly, nextPage's Navigation is different from other Navigation types, because it doesn't have metadata. So, if we want to be precise, we need to use another Navigation class for the nextPage link. I'd probably prefer to address it in the schema definition though, not here.

tests/test_components.py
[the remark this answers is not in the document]
VMRuiz: Case 2 is obviously need for: 
- Javascript pagination that is too complex to easily replicate with HTTP
- Filling input forms that load dynamically or contains sliders. For example, delivery food websites with location inputs that display a map for address selection, etc...   

In this case it would means: "Load page A and execute these XYZ actions to get to the next desired page" 

Case 1 is a lit bit more complex because I agree that the Page Objects should not be responsible of knowing if the next page would require rendering to fetch data properly. However, we still need some way to indicate to the crawler that page must be rendered. Maybe, that can be define as a dependency: 

- Page Object A define its WebPage must come from a rendering process
- Scrapy detects a Navigation with a URL + Callback that would trigger Page Object A 
- Scrapy generates a Request with rendering enabled

The alternative would be:

- Always enabled rendering by default  - maybe configurable by netlocs -  
- Crawl non-rendered version by default, and then retry from the Page Object that require rendering

Both options seems very inefficient

tests/test_components.py
VMRuiz: Yes, it should be base64 encoded.

tests/test_components.py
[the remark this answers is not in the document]
kmike: > However, we still need some way to indicate to the crawler that page must be rendered.

@VMRuiz if a page object needs to use browser to render the page and extract the data, it should use [BrowserHtml](https://web-poet.readthedocs.io/en/stable/api-reference.html#web_poet.page_inputs.browser.BrowserHtml) dependency instead of HttpResponse dependency.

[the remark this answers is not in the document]
VMRuiz: Great, but we still a way to tell Scrapy to use browser rendering to build the BrowserHtml object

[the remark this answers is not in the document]
kmike: Yes; it's a matter of defining a scrapy-poet provider for BrowserHtml. Probably, we should put the implementation of this provider to scrapy-zyte-api package. It's not implemented yet. See https://scrapy-poet.readthedocs.io/en/stable/providers.html (async def `__call__` should be supported, if I'm not mistaken).

zyte_common_items/items.py
lopuhin: >  I'd probably prefer to address it in the schema definition though, not here.

@kmike do you mean to change the schema to include probability for `nextPage`?

lopuhin: @kmike to be clear, are you discussing how to best organize the common items definitions, or how to change the schema to simplify it? In other words, do we have a constraint that they need to be exactly the same, or are you considering simplifications in commit items which don't have to be in the schema? I guess there would be an obvious tension here, where we don't want to expose fields we don't need in the schemas, which leads to a simpler single schema, but results in less re-use of code when you define multiple schemas.

zyte_common_items/components.py
VMRuiz: Added both HEAD and CONNECT

tests/test_components.py
VMRuiz: Fixed

tests/test_components.py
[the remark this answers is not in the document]
VMRuiz: So, do we all agree with naming this `Request` with the possibility of adding `actions` field on the future?

tests/test_components.py
```diff
@@ -84,3 +86,23 @@ def test_real_estate_area():
...
+
+def test_navigation():
+    Navigation(url="https://example.com")
+    headers = [
+        Header(name="Content-Type", value="application/x-www-form-urlencoded"),
+        Header(name="Host", value="foo.example"),
+    ]
+    Navigation(
+        name="Post Test",
+        url="https://example.com/test",
```
BurnzZ, line 104: Let's also add a test that would try out `_Url` conversion with `url_to_str`

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    """Represents navigations for a page of product listings on a e-commerce website"""
+
+    #: Main URL from which the data is extracted.
+    url: str = attrs.field(converter=url_to_str)
+
+    #: Name of the category/page with the product list.
+    #:
+    #: Format:
+    #:
+    #: - trimmed (no whitespace at the beginning or the end of the description string),
```
BurnzZ, line 729:
```suggestion
    #: - trimmed (no whitespace at the beginning or the end of the description string)
```

VMRuiz: Done

tests/test_components.py
VMRuiz: Done

tests/test_components.py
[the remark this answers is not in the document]
VMRuiz: Component `Navigation` is now `Request`

zyte_common_items/items.py
VMRuiz: What are the action points here? Do I need to change anything on the PR ?

VMRuiz:
Added docs

docs/reference/items.rst
```diff
@@ -26,6 +26,13 @@ Product list
    :members:
    :inherited-members:
 
+Product Navigation
+==================
```
Gallaecio, line 30:
```suggestion
Product navigation
==================
```

docs/reference/pages.rst
```diff
@@ -23,6 +23,14 @@ Product list
 .. autoclass:: zyte_common_items.ProductListPage(**kwargs)
    :show-inheritance:
 
+Product Navigation
```
Gallaecio, line 26:
```suggestion
Product navigation
```

tests/test_components.py
```diff
@@ -84,3 +88,30 @@ def test_real_estate_area():
...
         value=123.45, unitCode="SQMT", raw="123.45 square meters", areaType="LOT"
     )
+
+
+def test_headers():
+    Header(name="Connection", value="keep-alive")
+    Header(name="Content-Type", value="text/html; charset=utf-8")
+
+
+def test_navigation():
```
Gallaecio, line 98:
```suggestion
def test_request():
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
 
     #: Area in the raw format, as it appears on the website.
     raw: str
+
+
+@attrs.define(kw_only=True)
+class Header(Item):
+    """A http header value"""
```
Gallaecio, line 292:
```suggestion
    """An HTTP header"""
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+
+    #: Name of header
```
Gallaecio, line 294:
```suggestion
    #: Name of the header
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+    name: str
+
+    #: Value of the header
+    value: str
+
+
+@attrs.define(kw_only=True)
+class Request(Item):
+    """Describe a HTTP request to load a page"""
```
Gallaecio, line 303:
```suggestion
    """Describe a web request to load a page"""
```

Probably safer to stay away from HTTP wording on this one.

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+
+    #: Name of the page which describes the URL.
+    name: Optional[str] = None
```
Gallaecio, line 306:
```suggestion
    #: Name of the page being requested.
    name: Optional[str] = None
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+
+    #: HTTP URL
+    url: str = attrs.field(converter=url_to_str)
+
+    #: HTTP Method
+    #:
+    #: Values: "GET" "POST" "PUT" "DELETE" "OPTIONS" "TRACE" "PATCH" "HEAD" "CONNECT"
```
Gallaecio, line 313:
```suggestion
    #: HTTP method
```

Since we are supporting all HTTP method values, I don’t think we need to name it. We may even support arbitrary ones, right?

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+    method: str = "GET"
+
+    #: HTTP Request Body
+    #:
+    #: Format:
+    #: - Base64 Encoded
```
Gallaecio, line 319:
```suggestion
    #: HTTP request body, Base64-encoded
```

zyte_common_items/components.py
```diff
@@ -285,3 +285,42 @@ class RealEstateArea(Item):
...
+    body: Optional[str] = None
+
+    #: HTTP Headers
```
Gallaecio, line 322:
```suggestion
    #: HTTP headers
```

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
 
     #: Contains metadata about the data extraction process.
     metadata: Optional[Metadata] = None
+
+
+@attrs.define(kw_only=True)
+class ProductNavigation(Item):
+    """Represents navigations for a page of product listings on a e-commerce website"""
```
Gallaecio, line 720:
```suggestion
    """Represents the navigational aspects of a product listing page on an 
    e-commerce website"""
```

zyte_common_items/items.py
```diff
@@ -712,3 +713,38 @@ class RealEstate(Item):
...
+    #: - trimmed (no whitespace at the beginning or the end of the description string)
+    categoryName: Optional[str] = None
+
+    #: List of sub-category links ordered by their position in the page.
+    subCategories: Optional[List[Request]] = None
+
+    #: List of product links found on the page category ordered by their position in the page.
+    items: Optional[List[Request]] = None
+
+    #: The navigation to the next page, if available.
```
Gallaecio, line 738:
```suggestion
    #: A link to the next page, if available.
```

Gallaecio (commented):
Doc nitpicks!

zyte_common_items/components.py
VMRuiz: TIL

VMRuiz:
> Doc nitpicks!

Code updated!

Gallaecio (approved):
Once @kmike’s points are resolved, :heavy_check_mark: from me.

VMRuiz:
Hello @kmike , is there anything else I can do to move this forward?

zyte_common_items/items.py
kmike: Hey @VMRuiz! As the schema is not changed for now (no decision), we should change the item definition to follow the schema more precisely, i.e. no metadata for nextPage link.

zyte_common_items/items.py
VMRuiz: My two cents is that we should have a common Metadata class and it should be present on the schema of all attributes using Request.

Gallaecio: https://github.com/zytedata/zyte-common-items/pull/37/commits/a3f998f4961ede4aed8a5e3ee965d67ea4adc03c

zyte_common_items/items.py
Gallaecio: Is https://github.com/zytedata/zyte-common-items/pull/37/commits/d688ff783c649fcd69ed1c18945bc0a7dd4cdb88 a good solution for now?

kmike: @Gallaecio I think this solution is quite precise; it has an advantage of representing the schema correctly. The disadvantage is the weird developer API, when you need to be using ProbabilityRequest objects, which make very little sense for the use case of writing the extraction code.

What do you think about adding a default processor for the base page object classes, to allow users to return Request objects everywhere, and "promoting" them to ProbabilityRequest in the processor?

kmike: We may also be able to solve some of the issues with Metadata this way, e.g. by allowing users to return Metadata, and "demoting" it to classes with less fields if needed. In this case we can also add searchText to Metadata.

kmike: So, to summarize:

* For the developers we provide a single Metadata and a single Request class.
* Then, depending on the data type, we convert them to more precise types. We can do it in default processors. We may also duplicate it in attrs convertors, so that it works without processors as well.

What do you think?

zyte_common_items/pages.py
```diff
@@ -66,6 +74,10 @@ class BaseRealEstatePage(BasePage, Returns[RealEstate]):
     pass
 
 
+class BaseProductNavigationPage(BasePage, Returns[ProductNavigation]):
+    pass
```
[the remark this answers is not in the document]
Gallaecio: `ProductNavigation.metadata` is now `DateDownloadedMetadata`, which has no `probability` field.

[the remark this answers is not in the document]
kmike: Right, and if I'm not mistaken, BasePage sets it to a Metadata instance by default, which does have `probability` field.

Gallaecio:
Feels like I took a shortcut with https://github.com/zytedata/zyte-common-items/pull/37/commits/443deb3e9d217e7e15b5495fc15590007a9d17b2. I wonder if I should also create a `ProductFromListMetadata` type.

Creating a test to catch such scenarios, as I did for top-level items and pages, does not seem to make sense, though, since similar class  `ArticleFromList` replaces `metadata.probability` with `probability` as a direct attribute, and I imagine future “item-from-list” classes are likely to follow in its steps.

zyte_common_items/pages.py
```diff
@@ -1,35 +1,96 @@
...
+
+def _get_metadata_class(obj):
+    for base in getattr(obj.__class__, "__orig_bases__", []):
+        origin = getattr(base, "__origin__", None)
+        if not origin:
+            continue
+        if origin != HasMetadata:
+            continue
+        return _get_args(base)[0]
+    return None
```
kmike, line 62: With an implementation like this is makes sense to check that the inheritance works properly, e.g. a page object subclass which wants to use different metadata.

kmike:
> since similar class ArticleFromList replaces metadata.probability with probability as a direct attribute

This looks like a bug, I think the correct schema is with metadata.

kmike:
@Gallaecio could you please also check that we can make Request class work in a similar way?

Gallaecio:
> @Gallaecio could you please also check that we can make Request class work in a similar way?

How similar exactly? Just a renaming? 3 different classes? (1 for subcats, 1 for items, and the next page one) 2 per navigation item? 3 per navigation item?

kmike:
> How similar exactly? Just a renaming? 3 different classes? (1 for subcats, 1 for items, and the next page one) 2 per navigation item? 3 per navigation item?

I think it'd be nice to allow developers to return the same Request class both for nextPage and for items; that's the main thing. 

Unlike Metadata, there is no additional relevant information difference between Request and ProbabilityRequest in case of extraction written by developers; I'm not sure there are use cases for returning probability!=1 in metadata. So, I'd not bother with things like ProductNavigationNextPageReequest, all these requests look the same.

So, both options look fine: adding probability=1 when needed (i.e. auto-promoting to ProbabilityRequest, to get the correct export format), or dropping probability when needed (i.e. changing Request class to include probability, and having another class without probability).

zyte_common_items/pages.py
wRAR: I've added a test, can you please check that it tests what you wanted?

wRAR:
@kmike if we do this in a way similar to Metadata should we have a `HasRequest` base class similar to `HasMetadata` that will apply to types of `subCategories`, `items` and `nextPage` at the same time? As having separate base classes for these fields looks very complicated.

tests/test_items.py
```diff
@@ -615,3 +615,20 @@ def test_metadata():
...
         assert type(obj2.metadata) == metadata_cls
         assert obj2.metadata.dateDownloaded == "foo"
+
+
+def test_request():
+    """Test request class conversion on assignment"""
+
+    product_navigation = ProductNavigation(**_PRODUCT_NAVIGATION_MIN_KWARGS)
+    product_navigation.subCategories = [
+        Request(  # type: ignore[list-item]
```
kmike, line 625: If mypy doesn't like it here, it won't like it in the user code as well. Is there a way to make mypy happy?

tox.ini
```diff
@@ -36,4 +36,4 @@ basepython = python3.10
 deps =
     mypy==0.971
 
-commands = mypy zyte_common_items tests
+commands = mypy --show-error-codes zyte_common_items tests
```
kmike, line 39: 👍

zyte_common_items/components.py
```diff
@@ -426,6 +426,17 @@ class Video(_Media):
     """
 
 
+def cast_request(value: Request, cls: Type[Request]) -> Request:
+    new_value = copy_attributes(value, cls)
```
kmike, line 430: Is it always required to make a copy?

wRAR: No, we can skip this if the classes are the same. We don't do that for Metadata though, I guess we can add that there as well.

tests/test_items.py
wRAR: Tools don't know about `attrs` converters so this is expected, though I'm not sure why mypy doesn't consider other uses errors. In this case it's a more specific check about list items, not just e.g. `BasePage("http://example.com")`.

zyte_common_items/__init__.py
```diff
@@ -42,11 +52,15 @@
     BaseBusinessPlacePage,
     BasePage,
     BaseProductListPage,
+    BaseProductNavigationPage,
     BaseProductPage,
     BaseRealEstatePage,
     BusinessPlacePage,
+    HasMetadata,
+    MetadataT,
```
kmike, line 60: Why is it required to expose MetadataT?

zyte_common_items/util.py
```diff
@@ -69,3 +70,46 @@ def url_to_str(url: Union[str, _Url]) -> str:
 
 def format_datetime(dt):
     return f"{dt.isoformat(timespec='seconds')}Z"
+
+
+def copy_attributes(value: Any, new_cls: type) -> Any:
+    if type(value) == new_cls:
+        return value
```
kmike, line 77: As this function doesn't always copy, should it be renamed?

kmike: e.g. convert_to_class? Not sure

zyte_common_items/pages.py
kmike: In the test the subclass uses the same metadata class (ProductMetadata). I think it makes sense to test that it's possible to change the metadata class used in a subclass, e.g. use CustomProductMetadata in a ProductPage subclass.

kmike: Nevermind; I see now that the very next test checks exactly this :)

docs/reference/pages.rst
```diff
@@ -72,6 +80,10 @@ classes that rely on :class:`~web_poet.page_inputs.http.HttpResponse`.
 If you do not want :class:`~web_poet.page_inputs.http.HttpResponse` as input,
 you can inherit from :class:`~zyte_common_items.BasePage` instead.
 
+Your subclasses should also inherit generic classes
+:class:`web_poet.pages.Returns` and :class:`zyte_common_items.HasMetadata` to
```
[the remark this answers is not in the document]
wRAR: You can't write `Page[MyItem]` as `Page` is not a generic class.

You can change its declaration to `class Page(_BasePage[ItemT, MetadataT], WebPage[ItemT])` and then write `class ProductPage(Page[Product, ProductMetadata], HasMetadata[ProductMetadata])` (you can't omit `HasMetadata` as it adds `self.metadata_cls`) and the tests pass.

zyte_common_items/items.py
```diff
@@ -629,7 +642,7 @@ class ProductFromList(Item):
     mainImage: Optional[Image] = None
 
     #: Data extraction process metadata.
-    metadata: Optional[Metadata] = None
+    metadata: Optional[ProbabilityMetadata] = None
```
kmike, line 645: I think it'd be nice to have a converter here as well, so that passing Metadata works everywhere.

zyte_common_items/pages.py
```diff
@@ -1,35 +1,97 @@
...
+    ProductList,
+    ProductNavigation,
+    RealEstate,
+)
+from .util import format_datetime, metadata_processor
+
+try:
+    from typing import get_args as _get_args
+except ImportError:
+
```
kmike, line 32: could you please add a comment mentioning Python version, so that we don't forget to remove the workaround later?

zyte_common_items/util.py
```diff
@@ -69,3 +70,46 @@ def url_to_str(url: Union[str, _Url]) -> str:
...
+    input_attributes = {attribute.name for attribute in attrs.fields(value.__class__)}
+    output_attributes = {attribute.name for attribute in attrs.fields(new_cls)}
+    shared_attributes = input_attributes & output_attributes
+    new_value = new_cls(
+        **{attribute: getattr(value, attribute) for attribute in shared_attributes}
+    )
+    removed_nonempty_attributes = {
+        attribute
+        for attribute in (input_attributes - output_attributes)
+        if getattr(value, attribute) is not None
```
kmike, line 87: I think what "empty" means can vary. For example, removing probability=1 is fine if probability is not needed.

zyte_common_items/components.py
```diff
@@ -1,10 +1,109 @@
...
+    #: The probability (0 for 0%, 1 for 100%) that the resource features the
+    #: expected data type.
+    #:
+    #: For example, if the extraction of a product from a given URL is
+    #: requested, and that URL points to the webpage of a product with complete
+    #: certainty, the value should be `1`. If with complete certainty the
+    #: webpage features a job listing instead of a product, the value should be
+    #: `0`. When there is no complete certainty, the value could be anything in
+    #: between (e.g. `0.96`).
+    probability: Optional[float] = None
```
kmike, line 25: Shouldn't probability be 1 by default, not None?

zyte_common_items/components.py
```diff
@@ -1,10 +1,109 @@
...
```
kmike, line 55: Should we have it as 1 by default, not as None?

zyte_common_items/pages.py
wRAR: It's added in 3.8, I think we should just drop 3.7.

zyte_common_items/util.py
wRAR: Do you want to explicitly check and ignore the probability=1 case here? Or try to find if the field value is the default one?

kmike: I'm not sure what's the best implementation. I think we should ensure that if user passes Metadata instance as a value of some metadata attribute, then 

1) if probability is allowed for this attribute, then it's set to 1, unless user set it to a different value explicitly
2) if probability is not allowed, and user doesn't set probability explicitly when passing Metadata instance, there shouldn't be a warning
2) if probability is not allowed, but user sets it explicitly to a value which is not 1 (or 0? no sure about this), then a warning might be logged. 

I'm not 100% sure about (3), but I think we should ensure (1) and (2) work. My worry here was that if we add a default probability=1 to Metadata, then the current code will start showing warnings.

wRAR: I don't think we know here if the user passed the probability value explicitly? So we can only compare it with the default, and log the warning only if the value is not the default one? That's 2+3

kmike: What I mean is this:

```py
metadata=Metadata()
```

So, (1) means that metadata.probability should be 1 here; if probability=1 is used as a default value in Metadata.probability, that's achieved.

(2) means that if metadata attribute doesn't have probability, there shouldn't be a warning for in the example.

(3) implies a different code example:

```py
metadata=Metadata(probability=0.5)
```

^^ In this case it does make sense to have a warning. To address the warning, user should remove probability argument, converting it to the 1st example.

Then, there is a question on what to do in this case, when the probability is not supported:

```py
metadata=Metadata(probability=1)
```

I think it's fine both to ignore it (don't raise a warning), and to raise a warning.

wRAR: > I think it was the case before probability=1 was added as a default to Metadata class, but not the case now - without further work by addressing (1) we can break (2) .

Yes, assuming `MyItem.metadata` doesn't have `probability` (e.g. when `MyItem` is `ProductList`) we get a warning. How should we fix it? Add a check that if the removed attribute is `probability` and its value is `1.0` then don't count it for the warning?

wRAR: (Which was even easier than I thought, pushed)

zyte_common_items/items.py
```diff
@@ -81,9 +89,10 @@ class ArticleFromList(Item):
     #: All images.
     images: Optional[List[Image]] = None
 
-    #: The probability (0 for 0%, 1 for 100%) that this record represents an
-    #: actual article.
-    probability: Optional[float] = None
+    #: Data extraction process metadata.
+    metadata: Optional[ProbabilityMetadata] = attrs.field(
+        default=None, converter=attrs.converters.optional(MetadataCaster(ProbabilityMetadata)), kw_only=True  # type: ignore
```
kmike, line 94: Does the conversion work when list item metadata is accessed as a part of the `@field` in the page object, not as a part of the final item?

wRAR: `test_items.py::test_metadata_generic` accesses the field directly.

kmike: It checks that product_page.metadata field works correctly, but not product_list_page.products[0].metadata.

kmike: Yeah, looks good! I actually was a bit wrong asking this; my worry was that there is a convertor without a corresponding processor. But we don't have an Extractor defined for those list items, so there are no processors at all.

kmike:
Thanks @VMRuiz @wRAR @Gallaecio! 🚀

PR #38, opened 2023-05-02
Add ZyteKeepEmptyAdapter

wRAR:


zyte_common_items/adapter.py
```diff
@@ -99,3 +99,22 @@ def __iter__(self) -> Iterator:
...
+    to differentiate between empty and missing fields.
+    """
+
+    def __iter__(self) -> Iterator:
+        fields = [attr for attr in self._fields_dict if hasattr(self.item, attr)]
+        fields.extend(self.item._unknown_fields_dict)
+        return iter(fields)
+
+
+class ZyteKeepEmptyItemAdapter(ItemAdapter):
```
wRAR, line 117: I've added this and tests for it but haven't documented it yet. Initially it was created for easier test fixture creation but as it's now general purpose it looks like we either want `ItemAdapter` subclasses for both `ZyteItemAdapter` and `ZyteKeepEmptyAdapter` or for none of them, not just for the second one.

Also there is a serious naming problem here, mostly for an `ItemAdapter` subclass that enables `ZyteItemAdapter`.

wRAR: I've implemented most of this, should we also now rename ` ZyteItemAdapter` to `ZyteItemNoEmptyAdapter` and deprecate the old name?

docs/setup.rst
```diff
@@ -25,6 +26,17 @@ itemadapter.ItemAdapter.ADAPTER_CLASSES_ as early as possible in your code::
     ItemAdapter.ADAPTER_CLASSES.appendleft(ZyteItemAdapter)
 
 
+Alternatively, make your own subclass of :class:`itemadapter.ItemAdapter`::
+
+    from itemadapter import ItemAdapter
+    from zyte_common_items import ZyteItemAdapter
+
+    class MyItemAdapter(ItemAdapter):
+        ADAPTER_CLASSES = deque([ZyteItemAdapter]) + ItemAdapter.ADAPTER_CLASSES
```
kmike, line 35: You also need a deque import in this example.

See also: https://github.com/scrapy/itemadapter/issues/71

zyte_common_items/adapter.py
kmike: +0.2 from me to keep the ZyteItemAdapter name, to emphasize that's the recommended approach.

Gallaecio (approved):

PR #39, opened 2023-05-04
Add Article and ArticleList

Gallaecio:
To do:
- [x] Clarify the schema of `Article.authors[*]`.
- [x] Complete the implementation.


kmike (approved):
👍

zyte_common_items/items.py
```diff
@@ -20,10 +23,226 @@
...
+    #: In case there is no article data on the page or the page was not
+    #: reached, the returned "empty" item would still contain this URL field.
+    url: str = attrs.field(converter=url_to_str)
+
+    #: Data extraction process metadata.
+    metadata: Optional[DateDownloadedMetadata] = None
+
+
+@attrs.define(slots=True, kw_only=True)
+class ArticleList(Item):
```
kmike, line 199: If I'm not mistaken, we should keep only url, articles and metadata fields.

kmike (changes requested):
We should remove extra fields from the schemas.

zyte_common_items/items.py
Gallaecio: I’ve removed paginationNext and pageNumber], I think the rest may be meant to stay, even breadcrumbs.

PR #40, opened 2023-05-09
Add release notes for 0.5.0.

wRAR:


CHANGELOG.rst
```diff
@@ -2,6 +2,21 @@
...
 =========
 
+0.5.0 (YYYY-MM-DD)
+==================
+
+* Now requires ``itemadapter >= 0.8.0``.
+
+* Added :class:`~zyte_common_items.RealEstate`.
+
+* Added the :meth:`zyte_common_items.BasePage.no_item_found` method.
```
wRAR, line 12: This doesn't link to anything because the method is defined in `_BasePage` and we don't have an autoclass statement for it. Not sure if it's possible (and  preferable) to make Sphinx generate a doc for `BasePage.no_item_found()` based on the definition in the parent.

Also, `Page` inherits `_BasePage` directly, so technically we should either mention `_BasePage.no_item_found()` instead of `BasePage.no_item_found()`, or also mention `Page.no_item_found()`.

Thoughts?

VMRuiz: I found this

> autodoc_inherit_docstrings
> This value controls the docstrings inheritance. If set to True the docstring for classes or methods, if not explicitly set, is inherited from parents.
> 
> The default is True.
> 
> New in version 1.7.
> 

So it should be automatically added. Have you try generating the docs and checking if it does? 

> Also, Page inherits _BasePage directly, so technically we should either mention _BasePage.no_item_found() instead of BasePage.no_item_found(), or also mention Page.no_item_found().

I have no opinion about this. Maybe we could find examples on other libraries and see how this is approached?

wRAR: > Have you try generating the docs and checking if it does?

Well yes, that was th actual problem that it doesn't, but I've now checked the actual statements in pages.rst and Sphinx docs and it turns out we list `metadata` and `url` explicitly and don't use `:members:`, and Sphinx won't include members without `:members:`. So we just need to add a doc for `no_item_found()` to both public classes and link to them.

Gallaecio (approved):

PR #42, opened 2023-06-23
Breadcrumbs processor based on zyte-parsers

wRAR:


wRAR:
Do we need tests for pages themselves?

wRAR:
I also wonder what we should document about this and in which sections.

tests/test_processors.py
```diff
@@ -0,0 +1,46 @@
...
+    Breadcrumb(name="About", url="http://www.example.com/blog/about/"),
+]
+
+
+@pytest.mark.parametrize(
+    "input_value",
+    (
+        breadcrumbs_expected,
+        fromstring(breadcrumbs_html),
+        Selector(text=breadcrumbs_html),
```
kmike, line 30: I think it'd be nice to have a test for the "happy path" - a page object like this:

```py
class MyProductPage(ProductPage):
    @field
    def breadcrumbs(self):
        return self.css(".breadcrumbs")
```

If I'm not mistaken, it'd actually fail :) SelectorList vs Selector

kmike: Another scenario to test is returning a value which is different from the value which processor would extract, to ensure that the processor doesn't run.

zyte_common_items/processors.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .items import Breadcrumb
+
+
+def breadcrumbs_processor(value: Any, page: Any) -> Any:
+    def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
+        return Breadcrumb(name=value.name, url=value.url)
+
+    if isinstance(value, (Selector, HtmlElement)):
+        base_url = getattr(page, "url", None)
```
kmike, line 16: WebPage class also has base_url attribute

zyte_common_items/processors.py
```diff
@@ -0,0 +1,34 @@
...
+from zyte_parsers import extract_breadcrumbs
+
+from .items import Breadcrumb
+
+
+def breadcrumbs_processor(value: Any, page: Any) -> Any:
+    def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
+        return Breadcrumb(name=value.name, url=value.url)
+
+    if isinstance(value, SelectorList) and len(value) > 0:
```
kmike, line 15: Hm, either my comment disappeared, or I haven't posted it :) Shouldn't we return None if the SelectorList matches nothing? I think currently an empty SelectorList is going to be returned.

zyte_common_items/processors.py
wRAR: Do you mean we should use `base_url` if it's present and `url` otherwise? Or only `base_url`?

kmike: I'd actually do an isinstance check

zyte_common_items/processors.py
```diff
@@ -0,0 +1,42 @@
...
+    def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
+        return Breadcrumb(name=value.name, url=value.url)
+
+    if isinstance(value, SelectorList):
+        if len(value) == 0:
+            return None
+        value = value[0]
+
+    if isinstance(value, (Selector, HtmlElement)):
+        zp_breadcrumbs = extract_breadcrumbs(value, base_url=_get_base_url(page))
```
kmike, line 28: Could you please add a test for base url handling?

tests/test_processors.py
```diff
@@ -0,0 +1,68 @@
...
+        ("foo", "foo"),
+        (Selector(text="<html></html>"), None),
+        (SelectorList([]), None),
+        (breadcrumbs_expected, breadcrumbs_expected),
+        (fromstring(breadcrumbs_html), breadcrumbs_expected),
+        (Selector(text=breadcrumbs_html), breadcrumbs_expected),
+        (
+            [
+                zp_Breadcrumb(name="Home", url="http://www.example.com/blog/"),
+                zp_Breadcrumb(name="About", url="http://www.example.com/blog/about/"),
```
kmike, line 39: To ensure we cover the use case, could you please add a test for a page object which calls zyte_parsers.extract_breadcrumbs itself, and returns the result?

kmike:
The code looks good to me. We need some docs for the feature. Let's start with adding a docstring to the processor function.

zyte_common_items/processors.py
wRAR: Do you want a test that uses `ResponseShortcutsMixin.base_url` to get one from HTML? As we already test the simplest case with `BasePage.url`.

kmike: Yeah, make sure that if a page contains e.g. `<base>` tag, the right base url is used.

tests/test_processors.py
wRAR: We actually don't :)

wRAR:
I've added the docstring but it's never used in the docs. We also want some general words about processors, even though they are used by default. Now, where is the good place for that?

kmike:
👍

PR #46, opened 2023-07-11
Add brand_processor().

wRAR:


zyte_common_items/processors.py
```diff
@@ -49,3 +49,22 @@ def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
...
+    Other inputs are returned as is.
+    """
+
+    if isinstance(value, SelectorList):
+        if len(value) == 0:
+            return None
+        value = value[0]
+
+    if isinstance(value, (Selector, HtmlElement)):
+        return extract_brand_name(value, search_depth=2)
```
kmike, line 68: This seems to be a pattern which is going to be common for all processors. I'm fine with merging it as-is, when we only have 2 processors, but once we add 3rd, we should refactor it; no harm in refactoring it now as well.

kmike (approved):

zyte_common_items/processors.py
wRAR: Yeah.

Issue #49, opened 2023-07-19
HasMetadata doesn't work in some subclasses


wRAR:
This fails:

```python
def test_metadata_with_returns():
    class MyProduct(Product):
        pass

    class CustomProductPage(ProductPage, Returns[MyProduct]):
        pass

    url = ResponseUrl("https://example.com")
    html = b""
    page = CustomProductPage(response=HttpResponse(url=url, body=html))
    metadata = page.metadata
```

```
    @field
    def metadata(self) -> MetadataT:
>       value = self.metadata_cls()
E       TypeError: 'NoneType' object is not callable

zyte_common_items/pages.py:67: TypeError
```

`zyte_common_items.pages._get_metadata_class` in the case of this `CustomProductPage` cannot find any `HasMetadata` base and returns None.

wRAR:
AFAICS `__orig_bases__` is not recursive, so `_get_metadata_class` needs to be recursive?

AdrianAtZyte:
Fixed in #51, released in zyte-common-items 0.10.0 (2023-08-24).

PR #51, opened 2023-07-25
Fix _get_metadata_class to work recursively

wRAR:
A straightforward fix for #49 (at least assuming I've implemented BFS correctly), but I don't have a clear picture on what do `__orig_bases__` and `__origin__` do yet. Note that `__mro__` isn't suitable as it's depth-first.


Gallaecio (approved):

zyte_common_items/pages.py
```diff
@@ -44,18 +45,24 @@ class HasMetadata(Generic[MetadataT]):
...
-        if not origin:
-            continue
-        if origin != HasMetadata:
-            continue
-        return get_args(base)[0]
-    return None
+        return _get_metadata_class(type(self))
+
+
+def _get_metadata_class(cls: type) -> Type[MetadataT]:
```
kmike, line 51: Could you please a few more tests, for this function specifically, cvering the scenarios you considered while writing it?

kmike: I also wonder if HasMetadata could be made an argument, though no strong opinion

wRAR:
Requires new web-poet (https://github.com/scrapinghub/web-poet/pull/184)

setup.py
```diff
@@ -25,7 +25,7 @@
     install_requires=[
         "attrs>=21.3.0",
         "itemadapter>=0.8.0",
-        "web-poet>=0.7.0",
+        "web-poet>=0.14.0",
```
kmike, line 28: I think "min" environment in tox.ini needs to be updated according to this change. Would it make sense to add a comment to setup.py, to ensure we don't forget to do it with other version updates in install_requires?

PR #54, opened 2023-08-15
Add price processors.

wRAR:
TODO:

- [x] Refactor common code in all processors.
- [x] Refactor common code in BaseProductPage and ProductPage.
- [x] More use cases?


tests/test_pages.py
```diff
@@ -430,3 +436,151 @@ def nextPage(self):
...
+
+    html = b"""
+    <!DOCTYPE html>
+    <html>
+        <body>
+            <div>$13.2</div>
+        </body>
+    </html>
+    """
+    url = RequestUrl("https://example.com")
```
kmike, line 455: I think you can pass a string directly (though it won't probably allow to save lines of code here)

tests/test_pages.py
```diff
@@ -430,3 +436,151 @@ def nextPage(self):
...
+    page = CustomProductPage(response=HttpResponse(url=url, body=html))
+    assert page.price == "13.2"
+    assert page.currency is None
+    assert page.currencyRaw == "$"
+
+    # access currency fields before the price field
```
kmike, line 461: It may be nice to add a check which ensures the extraction code is not called multiple times; specifically, ensure the user defined code in the .price field is only executed once.

tests/test_pages.py
```diff
@@ -430,3 +436,151 @@ def nextPage(self):
...
+    page = CustomProductPage(response=HttpResponse(url=url, body=html))
+    assert page.currency is None
+    assert page.currencyRaw is None
+    assert page.price == "$13.2"
+
+
+def test_currency_hardcoded():
+    class CustomProductPage(ProductPage):
+        CURRENCY = "USD"
```
kmike, line 497: Should we have a basic test, which checks CURRENCY in isolation? You don't need to define the price field for it to work, right?

zyte_common_items/pages.py
```diff
@@ -133,9 +139,33 @@ class BaseJobPostingPage(
 
 
 class BaseProductPage(BasePage, Returns[Product], HasMetadata[ProductMetadata]):
+    _parsed_price: Optional[Price] = None
+
+    def _get_parsed_price(self):
+        if not hasattr(self, "price"):
+            return None
+        if self._parsed_price is None:
+            self.price
```
kmike, line 148: I think we should also support async price fields, and add tests for them

kmike: It seems it means that the currencyRaw field in the base class also becomes async 🙀

zyte_common_items/processors.py
```diff
@@ -59,12 +64,47 @@ def brand_processor(value: Any) -> Any:
...
+    """Convert the data into a price string if possible.
+
+    Supported inputs are :class:`~parsel.selector.Selector`,
+    :class:`~parsel.selector.SelectorList` and :class:`~lxml.html.HtmlElement`.
+    Other inputs are returned as is.
+    """
+
+    value = _handle_selectorlist(value)
+    if not isinstance(value, (Selector, HtmlElement)):
+        return value
```
kmike, line 105: This code also repeats a few times; maybe a time to define a decorator?

```py
@only_handle_selectorlist
def simple_price_processor(value: Any) -> Any:
    price = extract_price(value)
    if price.amount is None:
        return None
    return str(price.amount)
```

tests/test_pages.py
kmike, line 497: Could you please also add a test where user's page object defines both currencyRaw and price fields?

tests/test_pages.py
```diff
@@ -430,3 +436,151 @@ def nextPage(self):
         assert request.metadata is not None
         assert request.metadata.probability == 1.0
     assert type(page.nextPage) == Request
+
+
+def test_price_selector():
```
kmike, line 441: What do you think about moving all these price tests to a separate file? test_pages becomes large.

docs/usage/pages.rst
```diff
@@ -50,4 +50,12 @@ All ``breadcrumbs`` fields have the following processor enabled:
 
 .. autofunction:: zyte_common_items.processors.breadcrumbs_processor
 
+All ``price`` fields have the following processor enabled:
```
kmike, line 53: It seems the documentation should also mention the default imlementations for currency and currencyRaw fields (currenctly only metadata and url are mentioned).

kmike: It's a different kind of default implementation though (which does nothing out of box, but does some magic if user defines other things), but it still needs to be documented.

tests/test_pages.py
wRAR: It won't be called multiple times only if the user specified cached=True, because `hasattr(self, "price")` calls it again via `__get__` :-/

zyte_common_items/processors.py
wRAR: Is `value: str` a mistake here? Or what do you want the decorator to do? It looks like the decorated functions need to take `Union[Selector, HtmlElement]`.

kmike: ah, right, it should be a selector/html element; I got confused

tests/test_pages_price.py
```diff
@@ -0,0 +1,235 @@
...
+
+def test_currencyRaw_explicit():
+    class CustomProductPage(ProductPage):
+        @field
+        def price(self):
+            return self.css("div::text")
+
+        @field
+        def currencyRaw(self):
+            return "$"
```
kmike, line 215: I think the test should use a different currency raw symbol, because otherwise it doesn't check that the value comes from currencyRaw field, not from the HTML

tests/test_pages.py
kmike: Even if it doesn't work, we should have xfail test for it. 
But I think there must be a way to make it work :)

zyte_common_items/pages.py
```diff
@@ -54,6 +61,32 @@ def _get_metadata_class(cls: type) -> Optional[Type[MetadataT]]:
...
 
 
+class PriceMixin(FieldsMixin):
+    """Provides price-related field implementations."""
+
+    _parsed_price: Optional[Price] = None
+
+    async def _get_parsed_price(self):
+        if not hasattr(self, "price"):
+            return None
```
kmike, line 71: Is it the check which makes us call user's price field multiple times? If self.parsed_price is not None, we don't need to do this check anymore, right? 

And if it is None, would it work if you e.g. assing it to a varibale, or use getattr, to avoid accessing self.price two times here?

wRAR: I was able to reduce the calls but if you access `currencyRaw` and then access `price` it's still called twice, as the result of the first call isn't saved anywhere. I don't think we can do anything about it?

docs/reference/pages.rst
```diff
@@ -13,6 +13,17 @@ Product
 .. autoclass:: zyte_common_items.ProductPage(**kwargs)
    :show-inheritance:
 
+.. autoclass:: zyte_common_items.pages.PriceMixin
+
+   .. data:: currency
+      :type: str
+
+      Price currency ISO 4217` alphabetic code (e.g. ``"USD"``).
```
kmike, line 21: Is there a ` missing somewhere?

tests/test_pages_price.py
```diff
@@ -0,0 +1,244 @@
...
+    assert page.call_count == 1
+
+    # access currency fields before the price field
+    page = CustomProductPage(response=HttpResponse(url=url, body=html))
+    assert page.call_count == 0
+    assert page.currency is None
+    assert await page.currencyRaw == "$"
+    assert page.call_count == 1
+    assert page.price == "13.2"
+    assert page.call_count == 2
```
kmike, line 41: Even if the reason for this behavior is clear, I think it could make sense to mark this particular "2" as xfail, and open an issue in the repo. I'm fine with merging without addressing it now.

End-to-end the behavior doesn't make much sense; we might be calling price extraction multiple times when to_item is called.

tests/test_pages_price.py
```diff
@@ -0,0 +1,244 @@
...
+        <body>
+            <div>$13.2</div>
+        </body>
+    </html>
+    """
+    url = "https://example.com"
+    page = CustomProductPage(response=HttpResponse(url=url, body=html))
+    assert page.price == "$13.2"
+    assert page.currency is None
+    assert await page.currencyRaw is None
```
kmike, line 63: Would it make sense to check the call count in this test as well? The code path looks different enough.

zyte_common_items/pages.py
```diff
@@ -54,6 +61,33 @@ def _get_metadata_class(cls: type) -> Optional[Type[MetadataT]]:
...
+class PriceMixin(FieldsMixin):
+    """Provides price-related field implementations."""
+
+    _parsed_price: Optional[Price] = None
+
+    async def _get_parsed_price(self):
+        if self._parsed_price is None:
+            price_field = getattr(self, "price", None)
+            if price_field is None:
+                return None
```
kmike, line 73: Does it mean that if `price` field returns None, this None result is not cached, and price extraction will be called multiple times?

zyte_common_items/pages.py
```diff
@@ -54,6 +61,33 @@ def _get_metadata_class(cls: type) -> Optional[Type[MetadataT]]:
     return get_generic_param(cls, HasMetadata)
 
 
+class PriceMixin(FieldsMixin):
+    """Provides price-related field implementations."""
+
+    _parsed_price: Optional[Price] = None
+
+    async def _get_parsed_price(self):
```
kmike, line 69: Would it work if we use cached_method decorator, instead of using an explicit _parsed_price attribute? Or are there issues with slots, etc. which made you go with the custom route?

tests/test_pages_price.py
wRAR: After adding call count checks to `test_price_explicit` we now have 3 asserts in total that should be xfailed, but I'm not sure how to cleanly extract them into tests that can be marked with xfail without duplicating all of them.

zyte_common_items/pages.py
wRAR: `price_processor` needs to assign the value directly, so `cached_method` didn't look like it will be enough here.

zyte_common_items/pages.py
kmike: It seems worth doing, I don't see downsides in using a sentinel here.

tests/test_pages_price.py
```diff
@@ -0,0 +1,253 @@
...
+    assert page.call_count == 1
+
+    # access currency fields before the price field
+    page = CustomProductPage(response=HttpResponse(url=url, body=html))
+    assert page.call_count == 0
+    assert page.currency is None
+    assert await page.currencyRaw == "$"
+    assert page.call_count == 1
+    assert page.price == "13.2"
+    assert page.call_count == 2  # we want this to be 1
```
kmike, line 41: this approach looks fine

zyte_common_items/pages.py
```diff
@@ -54,6 +61,36 @@ def _get_metadata_class(cls: type) -> Optional[Type[MetadataT]]:
...
+                return self._parsed_price
+            # the field wasn't executed yet
+            await ensure_awaitable(price_field)
+        return self._parsed_price
+
+    @field
+    def currency(self) -> Optional[str]:
+        if hasattr(self, "CURRENCY"):
+            return self.CURRENCY
+        return None
```
Gallaecio, line 84: :lipstick: 

```suggestion
        return getattr(self, "CURRENCY", None)
```

I actually verified `tox -e py311` still passes, because I’ve seen weird things about getattr and hasattr happen when attrs is involved :sweat_smile:

tox.ini
```diff
@@ -49,7 +50,7 @@ commands =
 [testenv:mypy]
 basepython = python3.11
 deps =
-    mypy==0.971
+    mypy==1.4.1
```
Gallaecio, line 53: This is also frozen in line 11, where upgrading causes issues with `tox -e py` on 3.11 for me.

zyte_common_items/pages.py
```diff
@@ -54,6 +61,36 @@ def _get_metadata_class(cls: type) -> Optional[Type[MetadataT]]:
...
 
 
+class PriceMixin(FieldsMixin):
+    """Provides price-related field implementations."""
+
+    _parsed_price: Optional[Price] = None
+
+    async def _get_parsed_price(self) -> Optional[Price]:
+        if self._parsed_price is None:
+            price_field = getattr(self, "price", object())
```
Gallaecio, line 71: Tests also pass like this:

```suggestion
            price_field = getattr(self, "price", None)
```

Gallaecio (commented):
I think we need to clarify a few things in the docs:

- Explain that the price processors use price-parser, with a link to the docs (README) of the library.
- Cover `CURRENCY` in the reference documentation of PriceMixin.currency.
- Explain in the reference documentation of PriceMixin.currencyRaw how it relies on the price field using the price processor.

zyte_common_items/pages.py
wRAR: Oh, it looks like I've added two fixes for the same problem.

wRAR:
Looks like mypy has different output on different Python versions.

wRAR:
So, as mypy outputs `expected "Union[str, _Url, None]"` on older Python versions and `expected "str | _Url | None"` on newer ones, and we run the same mypy on all supported Python versions, I see the following non-ideal options:

- don't check the message, only the error code (as I think `pytest-mypy-testing` can't compare even message substrings)
- only run the test on one Python version
- run the test with older mypy so that it always outputs `Union[str, _Url, None]`

mypy also has a `--python-version` argument but I haven't used it so far and `pytest-mypy-testing` doesn't seem to allow passing it anyway.

kmike:
I think that's fine to run typing tests only on the latest Python version.

kmike:
Thanks @wRAR!

PR #57, opened 2023-08-25
Force 2 decimal places for prices

Gallaecio:


wRAR (approved):

tests/test_processors.py
```diff
@@ -151,3 +156,19 @@ def brand(self):
...
+
+@pytest.mark.parametrize(
+    "input_value,expected_value",
+    [
+        ("", None),
+        ("1", "1.00"),
+        ("1.2", "1.20"),
+        ("1.23", "1.23"),
+        # Fails due to . being considered a thousands separator in this case.
+        # ("1.234", "1.234"),
```
Gallaecio, line 169: I wonder if we should expose [decimal handling override](https://github.com/scrapinghub/price-parser#decimal-separator), maybe the way we allow overriding the currency (class variable).

PR #58, opened 2023-08-31
Add description processors and fields.

wRAR:


zyte_common_items/pages.py
```diff
@@ -92,6 +97,58 @@ async def currencyRaw(self) -> Optional[str]:
...
 
 
+class DescriptionMixin(FieldsMixin):
+    """Provides description and descriptionHtml field implementations."""
+
+    _description_html: Union[HtmlElement, str, None] = None
+    _description_str: Optional[str] = None
+
+    _description_default = False
+    _descriptionHtml_default = False
```
wRAR, line 107: these flags are just for avoiding recursive calls of default fields, I haven't found a cleaner check

Gallaecio (approved):

tests/test_pages_description.py
```diff
@@ -0,0 +1,170 @@
+import pytest
+from web_poet import HttpResponse, field
+
+from zyte_common_items import ProductPage
+
+HTML = b"""
+<!DOCTYPE html>
+<html>
+    <body>
+        <article><p>text</p><p>text2</p></article>
```
kmike, line 10: What do you think about changing it, so that it's not the same as the cleaned version - e.g. use divs instead of p elements?

tests/test_pages_description.py
```diff
@@ -0,0 +1,176 @@
...
+
+    page = CustomProductPage(response=HttpResponse(url=url, body=HTML))
+    assert await page.description == TEXT_CLEANED
+    assert await page.descriptionHtml == DESCR_HTML_CLEANED
+
+
+@pytest.mark.asyncio
+async def test_description_selector():
+    # description is extracted from the returned node
+    # descriptionHtml is extracted from the same node
```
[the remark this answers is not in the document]
kmike: i.e. change the test data so that `to_html(TEXT_CLEANED) != DESCR_HTML_CLEANED`

zyte_common_items/pages.py
```diff
@@ -92,6 +97,58 @@ async def currencyRaw(self) -> Optional[str]:
...
+
+    async def _get_description(self) -> Optional[str]:
+        if self._description_default:
+            return None
+        if self._description_str is None:
+            description = self.description
+            if isinstance(description, CoroutineType):
+                description = await description
```
kmike, line 115: What's the reason for not using ensure_awaitable?

zyte_common_items/pages.py
```diff
@@ -92,6 +97,58 @@ async def currencyRaw(self) -> Optional[str]:
...
+    _description_html: Union[HtmlElement, str, None] = None
+    _description_str: Optional[str] = None
+
+    _description_default = False
+    _descriptionHtml_default = False
+
+    async def _get_description(self) -> Optional[str]:
+        if self._description_default:
+            return None
+        if self._description_str is None:
```
kmike, line 112: Should we use another sentinel, not None, for the cases self.description returns None after a long computation?

zyte_common_items/util.py
```diff
@@ -100,3 +100,21 @@ def __init__(self, target):
 
     def __call__(self, value):
         return cast_metadata(value, self._target)
+
+
+def wrap_description_into_html(description: str) -> str:
```
kmike, line 105: I think this function is specific to a particular field; it might be better to move it to processors.py or some other location, not add it to the utils.py

zyte_common_items/util.py
```diff
@@ -100,3 +100,21 @@ def __init__(self, target):
...
+    The format tries to match clear_html.cleaned_node_to_html().
+
+    >>> wrap_description_into_html('')
+    '<article>\n\n</article>'
+    >>> wrap_description_into_html('foo')
+    '<article>\n\n<p>foo</p>\n\n</article>'
+    >>> wrap_description_into_html('foo\nbar')
+    '<article>\n\n<p>foo</p>\n\n<p>bar</p>\n\n</article>'
+    >>> wrap_description_into_html('foo\n\nbar')
+    '<article>\n\n<p>foo</p>\n\n<p>bar</p>\n\n</article>'
```
kmike, line 117: I think it needs a bit more tests, e.g. a case with more \n's in the text, a case with leading/trailing \n, or a case where some lines are separated by 1 newline, and some are by 2.

zyte_common_items/util.py
```diff
@@ -100,3 +100,21 @@ def __init__(self, target):
...
+    """
+    paras_wrapped = [f"\n<p>{para}</p>\n" for para in description.split("\n") if para]
```
kmike, line 119: shouldn't para be html escaped?

tests/test_pages_description.py
```diff
@@ -0,0 +1,186 @@
+import pytest
+from web_poet import HttpResponse, field
+
+from zyte_common_items import ProductPage
+
+HTML = b"""
+<!DOCTYPE html>
+<html>
+    <body>
+        <article><div>text</div><p style="color:blue">text2&gt;</p><div class="twitter-tweet">tw</div></article>
```
kmike, line 10: Just in case, could you please also add some non-ascii text here?

zyte_common_items/processors.py
```diff
@@ -129,3 +131,87 @@ def simple_price_processor(value: Union[Selector, HtmlElement], page: Any) -> An
...
+        return None
+    assert isinstance(value, HtmlElement)
+    cleaned_node = clean_node(value, _get_base_url(page))
+    cleaned_text = cleaned_node_to_text(cleaned_node)
+    page._description_node = cleaned_node
+    page._description_str = cleaned_text
+    return cleaned_text
+
+
+def wrap_description_into_html(description: str) -> str:
```
kmike, line 190: What do you think about making this function private for now? We might need to change the name later, or change the implementation, because a similar function could be used for other similar fields in other data types, which may not be "description".

kmike (approved):

PR #60, opened 2023-09-05
Add release notes for 0.11.0.

wRAR:


CHANGELOG.rst
```diff
@@ -2,6 +2,25 @@
...
+* Now requires ``clear-html >= 0.4.0``.
+
+* Added :func:`zyte_common_items.processors.description_processor` and enabled
+  it for the ``description`` fields.
+
+* Added :func:`zyte_common_items.processors.description_html_processor` and
+  enabled it for the ``descriptionHtml`` fields.
+
+* Added default implementations for the ``description`` and ``descriptionHtml``
+  fields (if one of these is user-defined, another one will use it).
```
kmike, line 17: Is it for ProductPage?

wRAR: Do you mean mentioning this explicitly? As we don't mention the same in the previous changelog entry, about currency fields.

kmike: The difference is that there are other data types which have description field, e.g. Article and BusinessPlace.

wRAR: #61

Gallaecio (approved):
Provided we merge https://github.com/zytedata/zyte-common-items/pull/61 as well, which seems to be the plan.

CHANGELOG.rst
wRAR: Right.

CHANGELOG.rst
```diff
@@ -2,6 +2,26 @@
...
+* Now requires ``clear-html >= 0.4.0``.
+
+* Added :func:`zyte_common_items.processors.description_processor` and enabled
+  it for the ``description`` fields (except ``Article.description``).
+
+* Added :func:`zyte_common_items.processors.description_html_processor` and
+  enabled it for the ``descriptionHtml`` fields.
+
+* Added default implementations for the ``description`` (except
+  ``Article.description``) and ``descriptionHtml`` fields (if one of these is
```
kmike, line 17: should it be ArticlePage and BaseArticlePage, not Article?

wRAR: Yeah, I'm using `Article` as a shorthand here and in the docs, because I don't want to list both classes :) I can change this.

Gallaecio: Maybe “article page classes” if you don’t want to mention both, but no strong opinion on this.

kmike: I'm fine with "article page classes". At the same time, it also looks good to be very explicit in the changelog. 

For example,  this entry

> Added :func:`zyte_common_items.processors.description_processor` and enabled
  it for the ``description`` fields (except ``Article.description``).

can become outdated / ambiguous if we add another data type which has description field. To understand the change, one must know the current set of page objects / items at the time of the release.

That's not a big deal, I'm fine with not addressing it :) Just explaining the nit-picking over the names .

PR #63, opened 2023-10-04
Provide page objects for easier overriding

Gallaecio:


Gallaecio:
@kmike Shall I go ahead with this approach for all other types?

wRAR (approved):

tests/test_pages.py
```diff
@@ -155,16 +155,17 @@ def validate_input(self):
 
 
 def test_page_pairs():
-    """For every page a base page, for every base page a page."""
+    """For every page a base page and an auto page, and the other way around."""
```
BurnzZ, line 158: Not sure if this is the right sentence fix.

```suggestion
    """For every base page and an auto page, and the other way around."""
```

BurnzZ (approved):

tests/test_pages.py
Gallaecio: The test checks that for every base page there is both a matching page and a matching auto page, that for every page there is a matching base page, and that for every auto page there is a matching base page.

I think my proposed wording is the shortest to express that. But maybe my first paragraph in this comment would be a better docstring?

BurnzZ: How about `"""For every page, a base page and an auto page. Vice versa."""`. I think I'm confused with `"... page a base page ..."

kmike (approved):

PR #64, opened 2023-10-09
helper method: get_probability()

BurnzZ:
This makes it _slightly_ easier to access the probability value in the spider. It also avoids the redundant code that checks if metadata is present in case the instance is a `Request` and not a `ProbabilityRequest`.

### DIFF
--- a/zyte_common_items/components.py
+++ b/zyte_common_items/components.py
@@ -455,7 +455,7 @@ class Request(Item):
             method=self.method or "GET",
             headers=header_list,
             body=self.body_bytes,
-            **kwargs
+            **kwargs,
         )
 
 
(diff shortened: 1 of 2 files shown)


zyte_common_items/components.py
```diff
@@ -485,6 +485,11 @@ class ProbabilityRequest(Request):
     #: Data extraction process metadata.
     metadata: Optional[ProbabilityMetadata] = None
 
+    @property
+    def probability(self) -> Optional[int]:
+        if metadata := getattr(self, "metadata", None):
+            return metadata.probability
```
kmike, line 491: I think it's good to have this helper. However, I wonder what's the best way to avoid confusion for the users:

* one may see request.probability, and try to set request.probability in their code
* one may see request.probability, and wonder why isn't it in the output

The first approach would be to add setter to the property. It solves the first issue, but not the second one.
So, maybe a better approach could be to create a method (get_probability), instead of a property?

BurnzZ: Makes sense. Updated in https://github.com/zytedata/zyte-common-items/pull/64/commits/70fc2d12a07bc7231341911dd0809a067cdf88ef

PR #65, opened 2023-10-23
new ProbabilityMixin class with get_probability() on items

BurnzZ:
Related to https://github.com/zytedata/zyte-common-items/pull/64, it would seem it's beneficial to extend `get_probability()` on item instances as well.

**Motivation:** In a spider callback, it's easier to do something like:

```python
def parse_item(self, response, item: Product):
    probability = item.get_probability()
    if probability and probability >= 0.1:
        yield item
```

than:

```python
def parse_item(self, response, item: Product):
    probability = None
    if metadata := getattr(item, "metadata", None):
        probability = metadata.probability

    if probability and probability >= 0.1:
        yield item
```

### DIFF
--- a/zyte_common_items/base.py
+++ b/zyte_common_items/base.py
@@ -51,8 +51,7 @@ def _extend_trail(trail: _Trail, key: Union[int, str]):
 class ProbabilityMixin:
     def get_probability(self) -> Optional[float]:
         if metadata := getattr(self, "metadata", None):
-            if probability := getattr(metadata, "probability", None):
-                return probability
+            return getattr(metadata, "probability", None)
         return None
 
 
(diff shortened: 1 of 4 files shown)


Gallaecio (approved):

zyte_common_items/base.py
```diff
@@ -48,7 +48,16 @@ def _extend_trail(trail: _Trail, key: Union[int, str]):
 
 
 @attrs.define
-class Item(_ItemBase):
+class ProbabilityMixin:
+    def get_probability(self) -> Optional[float]:
+        if metadata := getattr(self, "metadata", None):
+            if probability := getattr(metadata, "probability", None):
+                return probability
```
kmike, line 55: Wouldn't it return None when the probability is 0?

BurnzZ: Good catch! https://github.com/zytedata/zyte-common-items/pull/65/commits/e5cececd92943c248d45c355768762713c5df220

PR #67, opened 2023-11-08
Provide complete docstring coverage for the public API

Gallaecio:


zyte_common_items/util.py
```diff
@@ -5,7 +5,7 @@
 import attrs
 from web_poet.page_inputs.url import _Url
 
-# Caches the attribute names for attr.s classes
+#: Caches the attribute names for attr.s classes.
 CLASS_ATTRS: WeakKeyDictionary = WeakKeyDictionary()
```
kmike, line 9: I'm not sure we want this to be public

zyte_common_items/util.py
```diff
@@ -47,18 +47,26 @@ def split_dict(dict: Dict, key_pred: Callable[[Any], Any]) -> Tuple[Dict, Dict]:
...
     if not isinstance(url, (str, _Url)):
         raise ValueError(
-            f"{url!r} is neither a string nor an instance of RequestURL or ResponseURL."
+            f"{url!r} is neither a string nor an instance of RequestUrl or ResponseUrl."
         )
     return str(url)
 
 
 def format_datetime(dt):
+    """Returns the specified :class:`~datetime.datetime` object, assumed to be
```
kmike, line 61:
```suggestion
    """Return the specified :class:`~datetime.datetime` object, assumed to be
```

zyte_common_items/util.py
```diff
@@ -47,18 +47,26 @@ def split_dict(dict: Dict, key_pred: Callable[[Any], Any]) -> Tuple[Dict, Dict]:
...
+    in the UTC timezone, in ISO format, with the timezone specified as
+    ``Z``."""
     return f"{dt.isoformat(timespec='seconds')}Z"
 
 
 def convert_to_class(value: Any, new_cls: type) -> Any:
+    """Converts *value* into *type* keeping all shared attributes, and
```
kmike, line 68:
```suggestion
    """Convert *value* into *type* keeping all shared attributes, and
```

kmike (approved):

zyte_common_items/pages.py
```diff
@@ -231,33 +234,41 @@ def url(self) -> str:
...
+
     class Processors(BasePage.Processors):
         breadcrumbs = [breadcrumbs_processor]
 
 
 class BaseArticleNavigationPage(
     BasePage, Returns[ArticleNavigation], HasMetadata[ArticleNavigationMetadata]
 ):
-    pass
+    """:class:`BasePage` subclass for :class:`ArticleNavigation`."""
```
BurnzZ, line 255: Minor: In `components.py`, we added docstrings but left out `pass`. No strong opinion on which one to use but it'd be better to have consistency.

zyte_common_items/util.py
```diff
@@ -86,15 +94,23 @@ def convert_to_class(value: Any, new_cls: type) -> Any:
...
 
 
 def metadata_processor(metadata, page):
+    """Processor for a metadata field that ensures that the output metadata
+    object uses the metadata class declared by *page*."""
     return cast_metadata(metadata, page.metadata_cls)
 
 
 class MetadataCaster:
+    """attrs convertor that converts an input metadata object into the metadata
```
BurnzZ, line 111:
```suggestion
    """attrs converter that converts an input metadata object into the metadata
```

zyte_common_items/util.py
Gallaecio: Shall I prefix it with `_`?

PR #69, opened 2023-12-19
Provide extractors

Gallaecio:


tests/test_extractors.py
```diff
@@ -0,0 +1,99 @@
...
+
+        @field
+        def price(self):
+            return self.data["price"]
+
+        @field
+        def regularPrice(self):
+            return self.data["regularPrice"]
+
+    data = {"price": Selector("10€"), "regularPrice": Selector("20€")}
```
Gallaecio, line 30: I was expecting this to work without `Selector`, but it does not. Should we consider making some of the processors in https://github.com/zytedata/zyte-parsers able to handle strings as inputs?

wRAR: I've forgot what were our design decisions with respect to returning strings from fields. @kmike ?

Did we want to make users disable the processor in such cases?

wRAR: Actually the design decisions I was thinking about were for the processors in zyte-common-items, not for lower-level functions in zyte-parsers, where it should be find to add text input support where possible.

wRAR: The support in zyte-parsers is added in https://github.com/zytedata/zyte-parsers/pull/11, but this code uses `MyProductFromListExtractor` for which the question stilll stands.

wRAR (approved):

tests/test_extractors.py
```diff
@@ -0,0 +1,99 @@
...
```
BurnzZ, line 73: Let's have another case where another key exists in the dat input that is not declared as a `@field`.

Gallaecio: Could you elaborate on the aim of that? Would https://github.com/zytedata/zyte-common-items/pull/69/commits/3f8d115227f6e7a44cfca13a25f2ccf55988a867 work?

zyte_common_items/processors.py
```diff
@@ -100,8 +120,8 @@ def brand_processor(value: Union[Selector, HtmlElement], page: Any) -> Any:
     return extract_brand_name(value, search_depth=2)
 
 
-@only_handle_nodes
-def price_processor(value: Union[Selector, HtmlElement], page: Any) -> Any:
+@_handle_price
+def price_processor(value: Union[Selector, HtmlElement, str, None], page: Any) -> Any:
```
wRAR, line 124: @kmike do we want to pass the strings returned from the user-defined field to pass through or parse them?

tests/test_extractors.py
BurnzZ: I was thinking along the lines of the data being `{"price": "10€", "regularPrice": "20€", "other-data": "value"}` while the `@fields` are only `price` and `regularPrice`.

Gallaecio: https://github.com/zytedata/zyte-common-items/pull/69/commits/53a3c7378a84bd0d0a29aa03d16bfafc3fd4a3ec

zyte_common_items/processors.py
kmike: It seems it'd be good to have passthrough for a processor which is enabled by default. Currently all other processors do a passthrough for strings, so it seems it'd be good to follow this.

BurnzZ (approved):

zyte_common_items/processors.py
Gallaecio: So, when the input is JSON, should users [wrap strings with a `parsel.Selector`](https://github.com/zytedata/zyte-common-items/pull/69#discussion_r1432308724) to get the processor to work?

kmike: It seems it'd be good to have passthrough for a processor which is enabled by default. Currently all other processors do a passthrough for strings, so it seems it'd be good to follow this.

zyte_common_items/extractors.py
```diff
@@ -0,0 +1,39 @@
...
+
+    class Processors(_ProductProcessors):
+        pass
+
+
+class ProductVariantExtractor(Extractor[ProductVariant]):
+    """:class:`~web_poet.pages.Extractor` for :class:`ProductVariant`."""
+
+    class Processors(_ProductProcessors):
+        pass
```
wRAR, line 31:
```suggestion
        gtin = [gtin_processor]
```

zyte_common_items/processors.py
kmike: > So, when the input is JSON, should users https://github.com/zytedata/zyte-common-items/pull/69#discussion_r1432308724 to get the processor to work?

I think they can just call the price parsing function directly in this case.

kmike: We can also provide more aggressive, non-transparent processors, not enabled by default. But I'm not sure we should do it right now.

tests/test_pages_price.py
```diff
@@ -42,14 +42,14 @@ def price(self):
 
 
 @pytest.mark.asyncio
-async def test_price_explicit():
+async def test_price_string():
```
kmike, line 45: What's the reason for changing this test case?

kmike: Not the test case name (the new one is good), but the content?

zyte_common_items/processors.py
```diff
@@ -108,7 +108,7 @@ def brand_processor(value: Union[Selector, HtmlElement], page: Any) -> Any:
 
 
 @only_handle_nodes
-def price_processor(value: Union[Selector, HtmlElement], page: Any) -> Any:
+def price_processor(value: Union[Selector, HtmlElement, str, None], page: Any) -> Any:
```
wRAR, line 111: I don't think this can work? `@only_handle_nodes` calls the wrapped function only for `Union[Selector, HtmlElement]`.

wRAR (approved):

kmike (approved):

PR #71, opened 2024-01-16
Release notes for 0.14.0.

wRAR:


zyte_common_items/pages.py
```diff
@@ -410,6 +410,7 @@ class Processors(Page.Processors):
         breadcrumbs = [breadcrumbs_processor]
         description = [description_processor]
         descriptionHtml = [description_html_processor]
+        gtin = [gtin_processor]
```
wRAR, line 413: Missed this in the original PR. I wonder if we can deduplicate BaseFooPage.Processors and FooPage.Processors.

kmike (approved):

PR #73, opened 2024-01-25
Provide user-friendly field processor documentation

Gallaecio:


kmike (approved):

wRAR (approved):
When we discussed with @kmike that we need to expand this doc we decided that we also need to explain what inputs does each function supports more verbosely than the docstrings currently do, which is especially useful for `rating_processor()`. I don't know if it should be done by adding examples for all inputs for all functions or in some less verbose way, and we also don't need to do this in this PR, I'm just writing this so that the idea isn't lost.

Gallaecio:
For most processors I feel like giving an illustrative example of what they can do is enough. But for rating, with the ability to pass separate selectors for separate combinations of selectors, I think we should indeed provide more, although I would do that on the reference docs of the processor.

kmike:
> For most processors I feel like giving an illustrative example of what they can do is enough. In the end these processors aim to be black boxes where you point them to the raw data and they magically parse it as expected.

I think there are 2 approaches: 

* provide example of what processors can do, tell that they may handle other cases as well, and suggest to try, to see if they work
* provide examples for all the cases the processors can handle; if the website doesn't look like what's in the example, the user should expect that there is a high chance the processor won't work, and custom logic is needed.

It looks like the second approach may give slightly better experience for the user, and slightly more confidence when using the processors. But I guess it depends on how much the processors can't do, and what we know about these scenarios.

PR #75, opened 2024-02-05
Make the fields of Auto- classes sync

Gallaecio:


wRAR (approved):

tests/test_pages_auto.py
```diff
@@ -79,6 +79,13 @@
 )
 
 
+async def assert_expected_item(page, item):
+    assert await page.to_item() == item
+    # Ensure that all fields are sync.
+    for key in item.__slots__:
```
kmike, line 85: Is there an "attrs" way to get the list of fields? Or maybe itemadapter way?

kmike (approved):

tests/test_pages_auto.py
Gallaecio: :shrug:

Gallaecio: `ItemAdapter` uses `[a.name for a in attr.fields(item_class)]`, but I guess I can just use `ItemAdapter` itself, since it is a dependency already.

PR #77, opened 2024-02-06
Implement an item pipeline to convert back to the Zyte Automatic Extraction schema

Gallaecio:
I’m keeping this off the docs on purpose. Maybe we can cover it in the [migration docs](https://docs.zyte.com/zyte-api/migration/zyte/autoextract.html).


Gallaecio:
@kmike @lopuhin I am basically copying the classes from the main branch of zyte-autoextract, but they seem to be a subset of the [documented ones](https://docs.zyte.com/automatic-extraction/api.html), e.g. no `Product.hasVariants`. Which source should I trust? (cc @kishan3)

zyte_common_items/base.py
```diff
@@ -75,7 +76,7 @@ def from_dict(cls, item: Optional[Dict]):
     @classmethod
     def _from_dict(cls, item: Optional[Dict], *, trail: _Trail = None):
         """Read an item from a dictionary."""
-        if not item:
+        if item is None:
```
Gallaecio, line 79: So that `{}` turns into an empty instance of the corresponding type, instead of `None`.

zyte_common_items/base.py
```diff
@@ -126,9 +127,7 @@ def _apply_field_types_to_sub_fields(cls, item: Dict, trail: _Trail = None):
         """
         from_dict, from_list = {}, {}
 
-        annotations = ChainMap(
-            *(c.__annotations__ for c in cls.__mro__ if "__annotations__" in c.__dict__)
-        )
+        annotations = ChainMap(*(get_type_hints(c) for c in cls.__mro__))
```
Gallaecio, line 130: Got hit by [PEP-649](https://peps.python.org/pep-0649/) due to the forward reference in `AEProduct.hasVariants`, glad https://github.com/scrapinghub/scrapy-poet/pull/188 was fresh in my memory :slightly_smiling_face:

zyte_common_items/base.py
```diff
@@ -148,8 +147,12 @@ def _apply_field_types_to_sub_fields(cls, item: Dict, trail: _Trail = None):
                 origin = get_origin(type_annotation)
 
             if origin is list:
-                value = item.get(field)
-                if not isinstance(value, list) and not (is_optional and value is None):
+                value = item.get(field, _UNDEFINED)
```
Gallaecio, line 150: If a field is missing rather than `None`, do not fail because it is not a list.

lopuhin:
> I’ll start from autoextract-poet classes but update them according to the documentation specs.

thanks @Gallaecio yes the docs on zyte.com should be up to date, and the ultimate source of truth for AE schemas is in https://github.com/scrapinghub/xod-schemas/blob/master/schema-generator/src/results.avsc.yml

wRAR (approved):

kmike (approved):

PR #79, opened 2024-03-01
Ignore non-HtmlElement nodes when parsing descriptions

Gallaecio:
Reported by @seagatesoft.

### DIFF
--- a/tests/test_pages_description.py
+++ b/tests/test_pages_description.py
@@ -223,7 +223,7 @@ def test_description_simple(page_class: type):
     assert page.description == TEXT_CLEANED
 
 
-def test_node():
+def test_non_html_node():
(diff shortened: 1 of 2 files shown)


wRAR (approved):

Gallaecio:
@wRAR After I created this PR, I thought came to me: Could it be more helpful to raise an exception that indicates the unsupported) type of node received rather than silently ignoring it? The current PR could lead a developer to believe that things do not work (e.g. due to some bug), rather than they are passing the wrong selector. Specially since it might not be obvious that we are only picking the first selector of a `SelectorList` (which I am now wondering if it is the right approach for description processors).

wRAR:
Not 100% sure but it looks better to throw an exception, yeah.

PR #80, opened 2024-03-05
Search requests (templates approach)

Gallaecio:
Search request support implemented by returning items that work as templates to be rendered into actual requests with some arguments.

Uses jinja2 for template handling, no strong opinion about that.

This approach is meant to support the following syntax:

```python
# web-poet example implementation
search_request_template = get_item(
    "https://books.toscrape.com/",
    SearchRequestTemplate,
)
search_request = search_request_template.request(keyword="foo")
search_result = get_item(
    search_request.url,
    BookList,
)

# scrapy-poet
def parse_homepage(self, response: DummyResponse, search_request_template: SearchRequestTemplate):
    for keyword in self.keywords:
        yield search_request_template.request(keyword=keyword).to_scrapy(self.parse_search_results)
```

~This implementation is limited to the `url` field [in line with the current implementation of scrapy-zyte-api](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/6ef5ba5bfbe549fdd5f6184504168b359ddd2b73/scrapy_zyte_api/providers.py#L158).
Before adding new fields, we might want to support them in scrapy-zyte-api.~

~For an alternative, see https://github.com/zytedata/zyte-common-items/pull/81.~


zyte_common_items/items.py
```diff
@@ -1202,3 +1204,26 @@ class SocialMediaPost(Item):
     metadata: Optional[SocialMediaPostMetadata] = attrs.field(
         default=None, converter=attrs.converters.optional(MetadataCaster(SocialMediaPostMetadata)), kw_only=True  # type: ignore
     )
+
+
+@attrs.define(kw_only=True)
+class RequestTemplate:
+    filters: Dict[str, Callable]
```
Gallaecio, line 1211: I don’t like this, but I cannot think of a better approach to enable custom filtering functions for templates, which might be necessary to handle corner cases.

kmike: What do you think about waiting, until we get a real-world use case for this? Then we can decide how best to handle it. Having a dict of callables in the item looks wrong, it'd complicate serialization, caching, etc.

docs/usage/request-templates.rst
```diff
@@ -0,0 +1,109 @@
...
+
+    from web_poet import handle_urls
+    from zyte_common_items import SearchRequestTemplatePage
+
+
+    @handle_urls("quotes.toscrape.com")
+    class QuotesToScrapeComSearchRequestTemplatePage(SearchRequestTemplatePage):
+        @field
+        def url(self):
+            return "https://quotes.toscrape.com/tag/{{ keyword|lower|replace(' ', '-') }}/"
```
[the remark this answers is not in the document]
Gallaecio: Is what you have in mind [something like this](https://github.com/zytedata/zyte-common-items/pull/80/commits/67c30f4501a4a44bbae3526c3e66589d5192a84e)?

```python
@attrs.define
class DSLSearchRequestTemplatePage(SearchRequestTemplatePage):
    request_url: RequestUrl

    class Processors:
        url = [edit_request_url]

    @field
    def url(self):
        return {"add_query_params": {"search": "{keyword}"}}
```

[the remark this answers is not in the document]
kmike: Yes, something like this. I think in a non-test scenario you don't need a separate page, just a processor which 

* supports non-string output from the url field, and 
* returns strings as-is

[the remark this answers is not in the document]
Gallaecio: > you don't need a separate page

You mean the `page` parameter of the processor? It’s syntax sugar so you don’t need to include `"url": self.request_url` in the dict you return from the field.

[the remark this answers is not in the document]
Gallaecio: Yes, in the future we could even define the processor in `SearchRequestTemplatePage.Processors` (future default processors is the only reason I declared that class).

docs/usage/request-templates.rst
```diff
@@ -0,0 +1,159 @@
...
+
+After you :ref:`write a request template page object
+<custom-request-template-page>` for a website, you can get a request template
+item for that website and call its ``request`` method to build a request with
+specific parameters.
+
+For example, after implementing the page object :ref:`below
+<custom-request-template-page>`, you can get a
+:class:`~zyte_common_items.SearchRequestTemplate` object for
+https://quotes.toscrape.com/, and build a search request for a specific
```
kmike, line 22: It's unfortunate that our website doesn't allow to have a nice search example. I'd expect the most common case to be a search form which generates a GET request, with added `q={keyword}`, not a tag lookup.

Do you think it'd be possible to have a simpler example here? Maybe even not using a real website.

tests/test_items.py
```diff
@@ -769,12 +769,18 @@ def test_metadata():
 
     -   If the generic Metadata object is assigned as input, it gets converted
         into an object of the right, more specific metadata class.
+
+    Request templates are an exception.
```
kmike, line 773: I think that can be fine to incude meta to the request template as well.

zyte_common_items/items.py
```diff
@@ -1202,3 +1206,64 @@ class SocialMediaPost(Item):
...
+
+    #: :doc:`Jinja template <jinja:templates>` for :class:`Request.body
+    #: <zyte_common_items.Request.body>`.
+    #:
+    #: It must be a plain :class:`str`, not :class:`bytes` or a Base64-encoded
+    #: :class:`str`. Base64-encoding is done by :meth:`request` after rendering
+    #: this value as a Jinja template.
+    #:
+    #: Defining a non-UTF-8 body is not supported.
+    body: Optional[str] = None
```
lopuhin, line 1240: Would this still work ok if we want a form-encoded body (e.g. a search form which makes a POST request), or it's too uncommon?

Gallaecio: Do you have any particular concern in mind? I assume it would be supported as long as the request body remained UTF-8, even though with a pure template (as opposed to [a future DSL](https://github.com/zytedata/zyte-common-items/pull/80#discussion_r1514159312) to simplify some common/tricky scenarios) it could be a bit cumbersome, and I wonder if we would need to expose some special variable to the template, like a random string of some length, to allow generating a different multi-part boundary every time (if hard-coding a boundary were to lower success rates).

tests/test_items.py
kmike: For the meta, I'd keep both probability and dateDownloaded. 

Ity may be useful even if we use https://github.com/TeamHG-Memex/Formasaurus to generate the templates, as we'd have e.g. a probability of a search form being a search form, and probability of search input being a search input.

zyte_common_items/items.py
lopuhin: > Do you have any particular concern in mind? 

Mostly about ergonomics of constructing the correct payload, exactly the point you covered, thanks 👍

kmike (approved):
Looks great, thanks @Gallaecio!

PR #81, opened 2024-03-05
Search requests (no-templates approach)

Gallaecio:
Alternative to https://github.com/zytedata/zyte-common-items/pull/80 that does not rely on templates.

This approach is meant to support the following syntax:

```python
# web-poet example implementation
search_request = get_item(
    "https://books.toscrape.com/",
    SearchRequest,
    page_params={"keyword": "foo"},
)
search_result = get_item(
    search_request.url,
    BookList,
)

# scrapy-poet
def start_requests(self):
    for keyword in self.keywords:
        yield Request(
            "https://example.com",
            meta={
                "page_params": {
                    "keyword": keyword,
                }
            },
        )

def parse(self, response: DummyResponse, search_request: SearchRequest):
    yield search_request.to_scrapy(self.parse_search_results)
```

The main advantage I see over https://github.com/zytedata/zyte-common-items/pull/80 is that there is no limit in how to implement the building of a request (e.g. you can parse the request URL and build the search URL with a different logic depending on it), at the cost of worse readability for the most common scenarios. Although we can probably improve readability with default processors implementing template support if we want.

Also, and this is true for both #80 and #81, we could potentially support in the future a way to allow removing 1 callback from the implementation if we allow specifying `SearchRequest` in the request meta, and have the provider resolve the request, e.g.
```python
# web-poet example implementation
search_result = get_item(
    "https://books.toscrape.com/",
    BookList,
    request=SearchRequest,
    request_params={"keyword": "foo"},
)

# scrapy-poet
def start_requests(self):
    for keyword in self.keywords:
        yield Request(
            "https://example.com",
            meta={
                "poet_request": SearchRequest,
                "poet_request_params": {
                    "keyword": keyword,
                }
            },
        )
```

### DIFF
--- a/tests/test_request_templates.py
+++ b/tests/test_request_templates.py
@@ -1,4 +1,4 @@
-from urllib.parse import quote, quote_plus
+from urllib.parse import quote, quote_plus, urlparse
 
 import attrs
 import pytest
@@ -53,6 +53,19 @@ class UrlBasedSearchRequestPage(ItemPage[SearchRequest]):
         return f"{self.request_url}?search={quote(self.page_params['keyword'])}"
 
 
+@attrs.define
+class DynamicSearchRequestPage(ItemPage[SearchRequest]):
+    page_params: PageParams
+    request_url: RequestUrl
+
+    @field
+    def url(self):
+        if "/category/" in urlparse(str(self.request_url)).path:
+            return f"{self.request_url}?category_search={quote(self.page_params['keyword'])}"
+        else:
+            return f"{self.request_url}?search={quote(self.page_params['keyword'])}"
+
+
 @pytest.mark.parametrize(
     ("page", "inputs", "keyword", "url"),
     (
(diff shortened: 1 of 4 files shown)


kmike:
> (e.g. you can parse the request URL and build the search URL with a different logic depending on it)

Could you please elaborate on this? Why can't you use request URL in the template approach?

kmike:
> we could potentially support in the future a way to allow removing 1 callback 

I think having callback may actually be good, because it allows to change the request constructing logic without changing the spider. 

In some cases you might need to download and parse HTML, in some cases it's not required. For example (probably a rare one), there could be some tokens in the search form which you need to submit, and they need to be parsed from HTML.

This also allows us to have a default implementation: detect search form on a page using a library like https://github.com/TeamHG-Memex/Formasaurus, construct a request from this form. This use case always requires downloading the page with the search form.

When downloading is required, it seems the templating approach works better, because downloading only needs to be done once (assuming e.g. tokens don't change); it gets more tricky if you need to call a page object for each keyword.

Gallaecio:
> > (e.g. you can parse the request URL and build the search URL with a different logic depending on it)
> 
> Could you please elaborate on this? Why can't you use request URL in the template approach?

Yes, I woke up realizing the “dynamic” test case was not showing this, as what this more easily handles is when the request parameter (keyword) determine the URL building. I’ve updated the PR with a better test case for that. And it should be possible to achieve the same in #80, to be clear; I will add a similar scenario there.

PR #91, opened 2024-04-11
Add new pipeline DropLowProbabilityItemPipeline

PyExplorer:
This pipeline was implemented during the work on `zyte-spider-templates` and it was decided to move it to this repository as a general solution for dropping items with low probability.


kmike:
@PyExplorer make sure to setup pre-commit, as described in https://zyte-common-items.readthedocs.io/en/latest/contributing.html

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,43 @@ def __init__(self):
...
+        self.threshold = None
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        o = cls(crawler.stats)
+        return o
+
+    def get_threshold(self, item, spider):
+        return spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {}).get(
+            item.__class__.__name__, DEFAULT_ITEM_PROBABILITY_THRESHOLD
```
Gallaecio, line 80: `item.__class__.__name__` is not very reliable. Could we use [global_object_name](https://github.com/scrapy/scrapy/blob/48c5a8c98f545e35708a580e724d0b8e1ada5e6e/scrapy/utils/python.py#L309)(`item.__class__`)?

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,43 @@ def __init__(self):
...
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        o = cls(crawler.stats)
+        return o
+
+    def get_threshold(self, item, spider):
+        return spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {}).get(
+            global_object_name(item.__class__), DEFAULT_ITEM_PROBABILITY_THRESHOLD
+        )
```
Gallaecio, line 83: We could read this setting in `__init__` and pre-process all keys, so that if they are not `str` we use `global_object_name` on them. That way, in code, using the classes as keys instead of their import path will also work.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,43 @@ def __init__(self):
...
+
+    async def process_item(self, item, spider):
+        if not self.threshold:
+            self.threshold = self.get_threshold(item, spider)
```
Gallaecio, line 87: Doesn’t this mean that only the first item class threshold will be taken into account, if a spider returns 2+ types of items?

PyExplorer: Oh, yes, this was the initial idea. Is it possible that the spider can produce multiple types of items?

Gallaecio: > Is it possible that the spider can produce multiple types of items?

If we consider custom, user-made subclasses, definitely. But also I wonder if, in the future, we might support e.g. crawling product reviews in addition to product themselves. I think it might be better to assume it will happen eventually and be ready for it, if it does not complicate things too much.

PyExplorer: Ahh, got it, makes sense. Thanks!

setup.py
```diff
@@ -31,6 +31,7 @@
         "price-parser>=0.3.4",
         "web-poet>=0.14.0",
         "zyte-parsers>=0.5.0",
+        "scrapy>=2.10",
```
Gallaecio, line 34: Could we lower this? [`global_object_name` was added in 1.4.0](https://github.com/scrapy/scrapy/commit/e285b1d6c2aaa1fdfe788f1894b0196bc64d1be1), even if it was improved later.

Gallaecio: Also, we need to update [tox.ini](https://github.com/zytedata/zyte-common-items/blob/c326028bfa9b779832d512c0e2c7f29a4681e91a/tox.ini#L31-L42) accordingly to make sure we test the minimum supported version as well.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
 
     def process_item(self, item, spider):
         return ae.downgrade(item)
+
+
+class DropItem(Exception):
+    """Drop item from the item pipeline."""
+
+    pass
```
Gallaecio, line 65: I think this class need to be `scrapy.exceptions.DropItem`, otherwise the items would not actually be dropped and this custom exception would be raised instead.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
...
+
+    def __init__(self, crawler):
+        self.stats = crawler.stats
+        self.thresholds = {}
+        self.init_thresholds(crawler.spider)
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        o = cls(crawler)
+        return o
```
Gallaecio, line 79: :lipstick: 

```suggestion
        return cls(crawler)
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
...
+
+    def init_thresholds(self, spider):
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+        for item, threshold in thresholds_settings.items():
+            item_type = item if isinstance(item, str) else global_object_name(item)
```
Gallaecio, line 84: I’m thinking my suggestion to use `global_object_name` was misguided.

Using [`load_object`](https://github.com/scrapy/scrapy/blob/1d11ea3a54607b436f9a88f07911902a4882f0e8/scrapy/utils/misc.py#L53) if `str` may be better, as it later allows to implement `get_threshold` without `global_object_name` for better efficiency and readability.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
...
+            global_object_name(item.__class__), DEFAULT_ITEM_PROBABILITY_THRESHOLD
+        )
+
+    async def process_item(self, item, spider):
+        threshold = self.get_threshold(item, spider)
+
+        self.stats.inc_value("item/crawl/total", spider=spider)
+        item_proba = item.get_probability()
+        if item_proba is None or item_proba >= threshold:
+            self.stats.inc_value("item/crawl/extracted_with_high_proba", spider=spider)
```
Gallaecio, line 98: Have you thought of any alternative names for the stats? I wonder if something like `"item_probability_threshold/dropped"/"item_probability_threshold/kept"` may be best.

I also wonder if we really need `"item_probability_threshold/kept"`, as I expect its value to match `item_scraped_count`.

```suggestion
            self.stats.inc_value(, spider=spider)
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
...
+        for item, threshold in thresholds_settings.items():
+            item_type = item if isinstance(item, str) else global_object_name(item)
+            self.thresholds[item_type] = threshold
+
+    def get_threshold(self, item, spider):
+        return self.thresholds.get(
+            global_object_name(item.__class__), DEFAULT_ITEM_PROBABILITY_THRESHOLD
+        )
+
+    async def process_item(self, item, spider):
```
Gallaecio, line 92:
```suggestion
    def process_item(self, item, spider):
```

zyte_common_items/pipelines.py
Gallaecio: Hmm… Looking for existing stats about dropping stuff to keep the naming consistent, I found we already have a stat for this in the e-commerce spider: https://github.com/zytedata/zyte-spider-templates/blob/5f4c2c667d0a33029ed3ebcee1692d9ce61d98e4/zyte_spider_templates/spiders/ecommerce.py#L171

Maybe we could keep the same name here, and eventually modify the ecommerce spider to use this middleware instead of its own dropping logic, with no change to the stat name: `drop_item/product/low_probability`. But then we would need to have some mapping from known item classes to their Zyte API parameter name (i.e. `zyte_common_items.Product` → `product`). Maybe we can simply do `name = item.__class__.__name__; name[0] = name[0].lower()`.

zyte_common_items/pipelines.py
PyExplorer: Agree, wanted to avoid import scrapy, but we imported this anyway for utils

zyte_common_items/pipelines.py
```diff
@@ -53,3 +57,49 @@ def __init__(self):
...
+
+    pass
+
+
+class DropLowProbabilityItemPipeline:
+    """Drop item with a probability less than threshold."""
+
+    def __init__(self, crawler):
+        self.stats = crawler.stats
+        self.thresholds = {}
```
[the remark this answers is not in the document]
PyExplorer: I see some reasons to keep this here:
1. Define threshold as a class attribute and set this empty from the beginning
2. Populate this with `init_thresholds`
3. Avoid message from IDE "Instance attribute thresholds defined outside __init_" :) 

@Gallaecio do you think it's better to move this to init_threshold anyway?
Another option is just return dict from self.init_thresholds and set like
self.thresholds = self.init_thresholds(crawler.spider)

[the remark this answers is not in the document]
Gallaecio: I suspected reason 3 was behind this. It’s fine as it is :slightly_smiling_face:

zyte_common_items/pipelines.py
PyExplorer: Agree, nice function, trying to found something like that but couldn't. Thanks!

zyte_common_items/pipelines.py
PyExplorer: @Gallaecio just wondering - why do we need to use not async here?

Gallaecio: Is not that we need not to use async, is that we do not need to use async. There is no penalty to using either I think, my thinking here is that using `async` could be misleading, making people think it does async stuff. It also makes it harder to reuse this method from sync code. No strong opinion, though, it is OK to leave it as it is.

zyte_common_items/pipelines.py
PyExplorer: Haven't thought much about the names here, just left that I used in Articles. Using both stats (kept/dropped) I considered quite useful as they are close to each other in stats, it's clear where these metrics come from and easy to compare. You are right, in general case kept==scrapped, but if suddenly that are different - it can be helpful to start searching a bug.
Anyway, might be at this moment to have compatibility with e-commerce, we can keep only drop_item/{name}/low_probability.

zyte_common_items/pipelines.py
PyExplorer: Thanks, fixed already, you are right, let's leave it more clear without async.

setup.py
PyExplorer: Ah, this [global_object_name](https://github.com/scrapy/scrapy/commit/e285b1d6c2aaa1fdfe788f1894b0196bc64d1be1) is not used anymore - do we need to lower it(and fix tox) in this case? And if so - to which version?

Gallaecio: I think we should lower the version to [2.0.1](https://github.com/scrapy-plugins/scrapy-zyte-api/blob/e8b4b11f8b94edc5ccf29be076e4f7a118097bcb/setup.py#L27) at least, if we can.

Gallaecio: Interesting failure in CI, unrelated to your changes other than adding scrapy as a dependency. Let me look into that myself.

PyExplorer: 2.0.1 doesn't work - the test [test_components_request.py](https://github.com/zytedata/zyte-common-items/blob/main/tests/test_components_request.py#L65) fails. The issue relates to creating headers from multiple headers and I see that this issue was fixed in [2.7.0](https://docs.scrapy.org/en/latest/news.html#scrapy-2-7-0-2022-10-17) `Fixed Headers.getlist returning only the last header`.
I've checked 2.7.0 - all tests passed.  With 2.6.3 - the same issue. Left 2.7.0 as a min version.

Gallaecio:
@kmike How do you feel about depending on `scrapy`? I know we have tried to avoid that in the past. Should we keep it as an optional dependency?

zyte_common_items/pipelines.py
```diff
@@ -53,3 +58,44 @@ def __init__(self):
...
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+        for item, threshold in thresholds_settings.items():
+            item_type = load_object(item) if isinstance(item, str) else item
+            self.thresholds[item_type] = threshold
+
+    def get_threshold(self, item, spider):
+        return self.thresholds.get(type(item), DEFAULT_ITEM_PROBABILITY_THRESHOLD)
+
+    def get_item_name(self, item):
+        return item.__class__.__name__.lower()
```
Gallaecio, line 85: This would turn `ProductNavigation` into `productnavigation`, I think the goal should be for it to be `productNavigation` (i.e. match the Zyte API param name).

zyte_common_items/pipelines.py
```diff
@@ -53,3 +58,44 @@ def __init__(self):
...
+
+    def process_item(self, item, spider):
+        threshold = self.get_threshold(item, spider)
+
+        self.stats.inc_value("item/crawl/total", spider=spider)
```
Gallaecio, line 90: I don’t like this stat, at least not the naming. It sounds too generic, not specific to this pipeline, but it is counting the items that this pipeline sees, i.e. it is not counting items dropped by earlier item pipelines.

Maybe a prefix like `drop_low_probability_item_pipeline` would make sense, to make it clear where this setting comes from. Or maybe `probability_item_pipeline` if we rename the pipeline accordingly. So we could have `probability_item_pipeline/input_items` here (and `probability_item_pipeline/dropped_items/<type>` below; not sure if keeping the current e-commerce stat name is worth it).

PyExplorer: Ah, ok, if we are not going to keep e-commerce naming - yes, `probability_item_pipeline` sounds clear, might be even add word `drop` like `probability_drop_item_pipeline`.
And in this case 
probability_drop_item_pipeline/input_items/...
probability_drop_item_pipeline/dropped_items/...
And this `drop` directly adds the idea of this pipeline.

PyExplorer: but `probability_item_pipeline` is also good :)

setup.py
```diff
@@ -29,6 +29,7 @@
         "itemadapter>=0.8.0",
         "Jinja2>=2.7",
         "price-parser>=0.3.4",
+        "scrapy>=2.0.1",
```
kmike, line 32: @Gallaecio @PyExplorer I think we should keep this dependency optional. This looks easy to achieve, just by moving imports to methods which need them.

Gallaecio: OK, then I think we should add `extra` and `min-extra` tox environments that run with Scrapy installed.

zyte_common_items/pipelines.py
PyExplorer: @Gallaecio based on discussion [here](https://github.com/zytedata/zyte-common-items/pull/91#pullrequestreview-2004355909) and  if we are thinking about not using names like in e-commerce stats - what do you think about it?:
1. using this approach 
```
  self.stats.inc_value(
          f"probability_drop_item_pipeline/input_items/{self.get_item_name(item)}", spider=spider
      )
  item_proba = item.get_probability()
  if item_proba is None or item_proba >= threshold:
      self.stats.inc_value(
          f"probability_drop_item_pipeline/kept_items/{self.get_item_name(item)}", spider=spider
      )
      return item
  self.stats.inc_value(
          f"probability_drop_item_pipeline/dropped_items/{self.get_item_name(item)}", spider=spider
      )
```
2. Rename the pipeline to ProbabilityDropItemPipeline
cc @BurnzZ @wRAR @kmike

tests/test_components_request.py
```diff
@@ -64,12 +62,15 @@ def callback():
...
 
 
-@pytest.mark.skipif(
-    _SCRAPY_VERSION < Version("2.7.0"),
-    reason="https://github.com/scrapy/scrapy/issues/5515",
-)
 def test_request_to_scrapy_headers_with_the_same_name():
+    pytest.importorskip(
+        "scrapy",
+        minversion="2.7.0",
```
Gallaecio, line 68: TIL

tox.ini
```diff
@@ -38,10 +38,20 @@ deps =
...
+[testenv:extra]
+basepython = python3.8
+deps =
+    scrapy==2.0.1
+
+[testenv:min-extra]
+basepython = python3.8
+deps =
+    {[testenv:min]deps}
+    {[testenv:extra]deps}
```
Gallaecio, line 53:
```suggestion
[testenv:extra]
basepython = python3.8
deps =
    {[base]deps}
    scrapy

[testenv:min-extra]
basepython = python3.8
deps =
    {[testenv:min]deps}
    scrapy==2.0.1
```

Also, remember to remove `scrapy` from `base` above.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,49 @@ def __init__(self):
 
     def process_item(self, item, spider):
         return ae.downgrade(item)
+
+
+class DropLowProbabilityItemPipeline:
+    """Drop item with a probability less than threshold."""
```
BurnzZ, line 61:
```suggestion
    """Drop item when its probability is less than threshold."""
```

zyte_common_items/pipelines.py
BurnzZ: What about having the following?

- `low_proba_filter/processed`
- `low_proba_filter/dropped/Product`
- `low_proba_filter/dropped/Article`

It's short yet informative enough. I think _proba_ has a intuitive enough meaning similar to what `i` means in a for loop.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,49 @@ def __init__(self):
...
+        self.init_thresholds(crawler.spider)
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
```
BurnzZ, line 75: Let's document the purpose and usage for this setting so users can easily find it. Perhaps in the docstring of the Pipeline?

tox.ini
PyExplorer: @Gallaecio, I'm not quite sure about `remove scrapy from base above.` - I see that base doesn't have scrapy. Might be from here?
```
[testenv]
deps =
    {[base]deps}
    scrapy
```

zyte_common_items/pipelines.py
PyExplorer: Thanks @BurnzZ, but there is also another idea to use name of pipeline as a prefix to make this clear where the stats come from. Some suggestions [here](https://github.com/zytedata/zyte-common-items/pull/91#discussion_r1568603494) and [here](https://github.com/zytedata/zyte-common-items/pull/91#discussion_r1567791845)

tox.ini
Gallaecio: Yes, from there, my bad.

zyte_common_items/pipelines.py
Gallaecio: For the record, proba sounds weird to me. I would expect prob or probab when shortening probability.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
     def process_item(self, item, spider):
         return ae.downgrade(item)
+
+
+class DropLowProbabilityItemPipeline:
+    """Drop item when it's probability is less than the threshold.
+
+    This pipeline drops item if it's probability, defined in the settings,
+    is less than the specified threshold. Thresholds can be defined per item type
+    using `ITEM_PROBABILITY_THRESHOLDS` in two ways:
```
BurnzZ, line 65:
```suggestion
    """This pipeline drops an item if its probability, defined in the settings,
    is less than the specified threshold. 
    
    There are two ways that thresholds can be defined per item type using
    the `ITEM_PROBABILITY_THRESHOLDS` setting:
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+
+    def get_threshold(self, item, spider):
+        return self.thresholds.get(type(item), DEFAULT_ITEM_PROBABILITY_THRESHOLD)
+
+    def get_item_name(self, item):
+        item_name = item.__class__.__name__
+        return item_name[0].lower() + item_name[1:]
+
+    def process_item(self, item, spider):
+        import scrapy
```
Gallaecio, line 106: :lipstick: 

```suggestion
        from scrapy.exceptions import DropItem
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+
+        threshold = self.get_threshold(item, spider)
+
+        self.stats.inc_value("item/crawl/total", spider=spider)
+        item_proba = item.get_probability()
+        if item_proba is None or item_proba >= threshold:
+            return item
```
Gallaecio, line 113: Maybe we could do this to allow items with `None` as probability to be yielded if their threshold is 0?

```suggestion
        item_proba = item.get_probability() or 0
        if item_proba >= threshold:
            return item
```

Gallaecio (approved):
Looks good to me, I think we can merge as soon as we decide on the stat names.

zyte_common_items/pipelines.py
kmike: Should it be `or 1`?

zyte_common_items/pipelines.py
kmike: proba is prety common in ML community, e.g. scikit-learn uses it as a convention in method and argument names

zyte_common_items/pipelines.py
BurnzZ: The current setup is inline with the ecommerce setup: https://github.com/zytedata/zyte-spider-templates/blob/main/zyte_spider_templates/spiders/ecommerce.py#L168

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+        for item, threshold in thresholds_settings.items():
+            item_type = load_object(item) if isinstance(item, str) else item
+            self.thresholds[item_type] = threshold
+
+    def get_threshold(self, item, spider):
+        return self.thresholds.get(type(item), DEFAULT_ITEM_PROBABILITY_THRESHOLD)
+
+    def get_item_name(self, item):
+        item_name = item.__class__.__name__
+        return item_name[0].lower() + item_name[1:]
```
BurnzZ, line 103: Very minor: I think I prefer the current case since it accurately reflects the item class name.

zyte_common_items/pipelines.py
Gallaecio: Oh, I misunderstood the meaning of `None`, and I thought when `None` items were being filtered out, not returned :facepalm:

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+        item_proba = item.get_probability()
+        if item_proba is None or item_proba >= threshold:
+            return item
+
+        self.stats.inc_value(
+            f"drop_item/{self.get_item_name(item)}/low_probability", spider=spider
+        )
+        raise DropItem(
+            f"The item: {item!r} is dropped as the probability ({item_proba}) "
+            f"is below the threshold ({threshold})"
```
BurnzZ, line 120:
```suggestion
            f"This item is dropped since the probability ({item_proba}) "
            f"is below the threshold ({threshold}):\n{item!r}"
```
Minor: just so it's easier to read the logs as the sentence comes first containing the proba and threshold values before the actual item value which can be quite long. Otherwise, the reader might need to scroll down at the end to see such values.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+        for item, threshold in thresholds_settings.items():
+            item_type = load_object(item) if isinstance(item, str) else item
+            self.thresholds[item_type] = threshold
+
+    def get_threshold(self, item, spider):
```
BurnzZ, line 98:
```suggestion
    def get_threshold_for_item(self, item, spider):
```
minor: when reading the tests, I was expecting it to return the value set by DEFAULT_ITEM_PROBABILITY_THRESHOLD. Perhaps this can make it a bit clearer.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
     def process_item(self, item, spider):
         return ae.downgrade(item)
+
+
+class DropLowProbabilityItemPipeline:
+    """This pipeline drops an item if its probability, defined in the settings,
+    is less than the specified threshold.
+
+    There are two ways that thresholds can be defined per item type using
+    the `ITEM_PROBABILITY_THRESHOLDS` setting:
```
BurnzZ, line 65:
```suggestion
    There are two ways that the thresholds can be defined per item type using
    the ``ITEM_PROBABILITY_THRESHOLDS`` setting:
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,67 @@ def __init__(self):
...
+
+    With paths to Item classes:
+        ITEM_PROBABILITY_THRESHOLDS = {
+            "zyte_common_items.items.Article": 0.1,
+            "zyte_common_items.items.Product": 0.2
+        }
+
+    With Item classes:
+        from zyte_common_items import Article, Product
+        ITEM_PROBABILITY_THRESHOLDS = {
```
BurnzZ, line 75:
```suggestion
        from zyte_common_items import Article, Product
        
        ITEM_PROBABILITY_THRESHOLDS = {
```

zyte_common_items/pipelines.py
Gallaecio: We would have an inconsistency with the current e-commerce stat, and with scrapy-zyte-api (`scrapy-zyte-api/request_args/product`). But we all seem to agree on not trying to keep the current e-commerce stat, and since this is about items and not Zyte API params, the item class may indeed be the way to go.

However, from that point of view, I would expect the import path, i.e. 2 classes named `Product` can be 2 different classes.

zyte_common_items/pipelines.py
BurnzZ: After our standup call, here's what we have agreed on:

```
drop_low_probability_item/processed
drop_low_probability_item/processed/Article
drop_low_probability_item/processed/Product
drop_low_probability_item/kept
drop_low_probability_item/kept/Article
drop_low_probability_item/kept/Product
drop_low_probability_item/dropped
drop_low_probability_item/dropped/Article
drop_low_probability_item/dropped/Product
```

BurnzZ (approved):
LGTM! 🙌

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38,py39,py310,py311,py312,mypy,docs,twinecheck,min
+envlist = min,min-extra,py38,py39,py310,py311,py312,extra,mypy,docs,twinecheck
```
kmike, line 2: I think we should include new environments to github actions, so that they're executed on pull requests

zyte_common_items/pipelines.py
```diff
@@ -2,6 +2,8 @@
 
 from zyte_common_items import ae
 
+DEFAULT_ITEM_PROBABILITY_THRESHOLD = 0.1
```
kmike, line 5: Making it a pipeline attribute would make it easier to customize a pipeline - inherit from it, change a value.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,78 @@ def __init__(self):
...
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+        for item, threshold in thresholds_settings.items():
+            item_type = load_object(item) if isinstance(item, str) else item
+            self.thresholds[item_type] = threshold
+
+    def get_threshold_for_item(self, item, spider):
+        return self.thresholds.get(type(item), DEFAULT_ITEM_PROBABILITY_THRESHOLD)
```
kmike, line 100: What do you think about allowing user to set this threshold via `ITEM_PROBABILITY_THRESHOLDS[None]` or `ITEM_PROBABILITY_THRESHOLDS["default"]`? 

For example, it'd allow to use a more aggressive threshld for everything by default (e.g. 0.5), or disable default thresholding completely, only enabling it for specific item types (use 0 value for default).

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,78 @@ def __init__(self):
...
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+        for item, threshold in thresholds_settings.items():
+            item_type = load_object(item) if isinstance(item, str) else item
```
kmike, line 96: I think load_object already does this check, with some additional validation

```suggestion
            item_type = load_object(item)
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +55,78 @@ def __init__(self):
...
+        return item.__class__.__name__
+
+    def process_item(self, item, spider):
+        from scrapy.exceptions import DropItem
+
+        item_name = self.get_item_name(item)
+        item_proba = item.get_probability()
+        threshold = self.get_threshold_for_item(item, spider)
+
+        self.stats.inc_value("drop_low_probability_item/processed", spider=spider)
```
kmike, line 112: I _think_ all these spider=spider can be dropped; I have zero idea why do we have this parameter in StatsCollector :) Probably, it's a leftover from times where Crawler supported multiple spiders.

PyExplorer: Oh, thanks, TIL!

zyte_common_items/pipelines.py
```diff
@@ -53,3 +53,74 @@ def __init__(self):
...
+    With Item classes:
+        from zyte_common_items import Article, Product
+
+        ITEM_PROBABILITY_THRESHOLDS = {
+            Article: 0.1,
+            Product: 0.2
+        }
+    """
+
+    _DEFAULT_ITEM_PROBABILITY_THRESHOLD = 0.1
```
kmike, line 80: let's make it pubic (without underscore), as it's a useful attribute to override in a subclass

zyte_common_items/pipelines.py
PyExplorer: Yes, that's a good idea, fixed!

zyte_common_items/pipelines.py
PyExplorer: Revert this back - up to Scrapy 2.4 it didn't do this check.
https://github.com/scrapy/scrapy/blob/2.3/scrapy/utils/misc.py#L38
And we have a min version in tox for extra 2.0.1.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +53,82 @@ def __init__(self):
...
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+
+        self.default_threshold = (
+            thresholds_settings.pop(None, None)
+            or thresholds_settings.pop("default", None)
+            or self.DEFAULT_ITEM_PROBABILITY_THRESHOLD
```
kmike, line 100:
```suggestion
            or thresholds_settings.pop("default", self.DEFAULT_ITEM_PROBABILITY_THRESHOLD)
```
With the previous code it's not possible to set 0 as a default, as it'd fall back to self.DEFAULT_ITEM_PROBABILITY_THRESHOLD

kmike, line 100: "default" key, as well as the default value when no default is provided (0.1) need to be documented.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +53,82 @@ def __init__(self):
 
     def process_item(self, item, spider):
         return ae.downgrade(item)
+
+
+class DropLowProbabilityItemPipeline:
+    """This pipeline drops an item if its probability, defined in the settings,
+    is less than the specified threshold.
+
+    There are two ways that the thresholds can be defined per item type using
```
kmike, line 62: These approaches can also be mixed. I think we can have a single example, something like this:

```py
from zyte_common_items import Article

ITEM_PROBABILITY_THRESHOLDS = {
    Article: 0.2,
    "zyte_common_items.items.Product": 0.3,
    "default": 0.15,  # if not provided, 0.1 is used
}
```

zyte_common_items/pipelines.py
kmike: 👍 Could you please add docs for it?

zyte_common_items/pipelines.py
```diff
@@ -53,3 +53,82 @@ def __init__(self):
...
+        return cls(crawler)
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
+
+        self.default_threshold = (
+            thresholds_settings.pop(None, None)
+            or thresholds_settings.pop("default", None)
```
kmike, line 99: I think it's better to support only one of the constants ("default"?)

zyte_common_items/pipelines.py
```diff
@@ -53,3 +53,82 @@ def __init__(self):
...
+        self.init_thresholds(crawler.spider)
+
+    @classmethod
+    def from_crawler(cls, crawler):
+        return cls(crawler)
+
+    def init_thresholds(self, spider):
+        from scrapy.utils.misc import load_object
+
+        thresholds_settings = spider.settings.get("ITEM_PROBABILITY_THRESHOLDS", {})
```
kmike, line 95: I wonder if we should make a copy here. Later thresholds_settings.pop is used, and it's not clear to me if it is problematic or not. It modifies settings in place, and I wonder if it breaks e.g. multiple spiders in the same process (they may share the same settings - Setting objects are going to be different, but the exact dict as a setting value could be shared - I'm not sure). It may also slightly break logging of settings ("default" will be missing), if for some reason it happens after init_thresholds method is executed.

PyExplorer: Oh, never thought that the settings can be share between spiders, but sure, let's make this much more reliable with deepcopy

.github/workflows/test.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         - python-version: '3.12'
         - python-version: '3.8'
           toxenv: min
+        - python-version: '3.8'
+          toxenv: extra
```
kmike, line 28: What do you think about running "extra" on 3.12? 
For example, a new Scrapy could be released which drops support for 3.8; in this case "extra" might have to install older Scrapy, and we'll never check on the latest Scrapy.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +54,87 @@ def __init__(self):
...
+        from zyte_common_items import Article
+
+        ...
+
+
+        ITEM_PROBABILITY_THRESHOLDS = {
+            Article: 0.2,
+            "zyte_common_items.items.Product": 0.3,
+            "default": 0.15,  # if not provided, DEFAULT_ITEM_PROBABILITY_THRESHOLD is used
+        }
```
kmike, line 82:
```suggestion
        from zyte_common_items import Article

        ITEM_PROBABILITY_THRESHOLDS = {
            Article: 0.2,
            "zyte_common_items.Product": 0.3,
            "default": 0.15,  # if not provided, DEFAULT_ITEM_PROBABILITY_THRESHOLD is used
        }
```

zyte_common_items/pipelines.py
```diff
@@ -53,3 +54,87 @@ def __init__(self):
...
+        ...
+    """
+
+    DEFAULT_ITEM_PROBABILITY_THRESHOLD = 0.1
```
kmike, line 86: 💄 Just DEFAULT_THRESHOLD? The pipeline is already named DropLowProbabilityItemPipeline, so shorter name could be ok.

zyte_common_items/pipelines.py
```diff
@@ -53,3 +54,87 @@ def __init__(self):
...
+
+class DropLowProbabilityItemPipeline:
+    """This pipeline drops an item if its probability, defined in the settings,
+    is less than the specified threshold.
+
+    There are two ways that the thresholds can be defined per item type using
+    the ``ITEM_PROBABILITY_THRESHOLDS`` setting
+    - with paths to Item classes,
+    - with Item classes.
+    Additionally, for Item's classes without their own threshold, you can set a default value.
```
kmike, line 67:
```suggestion
    Additionally, for Item's classes without their own threshold, you can set a default value,
    using the "default" key.
```

kmike, line 67:
```suggestion
    By default, 0.1 threshold is used, i.e. items with probabillity < 0.1 are dropped.
    Thresholds can also be defined per item type using
    the ``ITEM_PROBABILITY_THRESHOLDS`` setting
    - with paths to Item classes,
    - with Item classes.
    The default value can be changed using "default" key in 
    ``ITEM_PROBABILITY_THRESHOLDS``.
```

PyExplorer: @kmike I've prepared the version for description - is it ok (with some changes from yours), if your variant to use as  final version?
```
   This pipeline drops an item if its probability, defined in the settings,
    is less than the specified threshold.

    The threshold can be set using the ITEM_PROBABILITY_THRESHOLDS setting or the DEFAULT_THRESHOLD
    class constant.

    Using ITEM_PROBABILITY_THRESHOLDS offers greater flexibility, allowing you to define thresholds
    for each Item class separately or set a default threshold for all other item classes.

    Thresholds for Item classes can be defined using either the path to the Item class or
    directly using the Item classes themselves.

    If a default value (key "default") is provided in ITEM_PROBABILITY_THRESHOLDS, it will be applied
    to all Item classes without their own specified thresholds. If no default value is provided,
    the threshold for such Item classes will default to the value set in DEFAULT_THRESHOLD,
    which is 0.1 by default i.e. items with probabillity < 0.1 are dropped.

    DEFAULT_ITEM_PROBABILITY_THRESHOLD will also be used if no ITEM_PROBABILITY_THRESHOLDS setting is provided.

    The example of using ITEM_PROBABILITY_THRESHOLDS:

        from zyte_common_items import Article

        ITEM_PROBABILITY_THRESHOLDS = {
            Article: 0.2,
            "zyte_common_items.Product": 0.3,
            "default": 0.15
        }
```

kmike: I think it's good to document the default behavior in the beginning, i.e. state very early that by default, 0.1 is used ("By default, 0.1 threshold is used, i.e. items with probabillity < 0.1 are dropped."). This is what happens if user enables the pipeline, without adding any configuration, and it's actualy a very reasonable default. 

I wouldn't mention DEFAULT_THRESHOLD class constant; this is very unlikely to be relevant for the users. 

> Thresholds for Item classes can be defined using either the path to the Item class or
    directly using the Item classes themselves.

👍 

> "default": 0.15

removing comment is fine, as it's documented early

kmike: > If a default value (key "default") is provided in ITEM_PROBABILITY_THRESHOLDS, it will be applied
    to all Item classes without their own specified thresholds. If no default value is provided,
    the threshold for such Item classes will default to the value set in DEFAULT_THRESHOLD,
    which is 0.1 by default i.e. items with probabillity < 0.1 are dropped.

I feel it's a bit too verbose. E.g. "it will be applied to all Item classes without their own specified thresholds" doesn't seem required (users understand what default is). "If no default value is provided,
    the threshold for such Item classes will default to the value set in DEFAULT_THRESHOLD,
    which is 0.1 by default i.e. items with probabillity < 0.1 are dropped." this also doesn't seem required, if we document default threshold early, and drop DEFAULT_THRESHOLD mentions.

.github/workflows/test.yml
PyExplorer: yes, makes sense, I've added this as additional requirement.

.github/workflows/test.yml
```diff
@@ -24,6 +24,11 @@ jobs:
         - python-version: '3.12'
         - python-version: '3.8'
           toxenv: min
+        - python-version: '3.8'
```
kmike, line 27: this is duplicate

zyte_common_items/pipelines.py
PyExplorer: Ok ,git it - this is a new variant:
```
This pipeline drops an item if its probability, defined in the settings,
    is less than the specified threshold.

    By default, 0.1 threshold is used, i.e. items with probabillity < 0.1 are dropped.

    You can customize the thresholds by using the ITEM_PROBABILITY_THRESHOLDS setting that offers
    greater flexibility, allowing you to define thresholds for each Item class separately or
    set a default threshold for all other item classes.

    Thresholds for Item classes can be defined using either the path to the Item class or
    directly using the Item classes themselves.

    The example of using ITEM_PROBABILITY_THRESHOLDS:

        from zyte_common_items import Article

        ITEM_PROBABILITY_THRESHOLDS = {
            Article: 0.2,
            "zyte_common_items.Product": 0.3,
            "default": 0.15
        }
```

kmike: Looks good!

tests/test_pipelines.py
```diff
@@ -0,0 +1,227 @@
...
+        ),  # default threshold is taken from settings with the key 'default' == 0
+    ],
+)
+def test_init_thresholds(
+    thresholds_settings,
+    default_threshold,
+    expected_thresholds,
+    expected_default_thresholds,
+):
+    scrapy = pytest.importorskip("scrapy")  # noqa
```
kmike, line 67: What do you think about moving it to the module level?

kmike (approved):
Looks good @PyExplorer, thanks for your patience!

I have one remaining comment, but I'm fine with merging without addressing it. It's about tests. The current tests in the PR check every pipeline method, and use a lot of mocking. 

I think this is a bit excessive; these methods (like get_item_name) look more like implementation details. In my experience, together with mock usage, this could make refactoring harder, not easier. Also, it means that we actually don't check the actual pipeline usage, with a crawler, spider, etc. 

I recall refactoring a couple of projects where I used this approach, and it was a nightmare :) It took me more time to update tests, and it was just a mundane effort, not something which helps to ensure code doesn't break. It would have been easier to just remove all tests, change the implementation, and write new tests. I'm not saying it's that bad here (not at all), but still, as a general approach, this is risky.

Because of this, I highly prefer tests which check the user-facing behavior, and use as little mocks as possible (usually, zero). For example, it's usually possible to create a real Crawler instance.

Some good examples:

* https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/tests/test_middlewares.py
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_ae_pipeline.py

tests/test_pipelines.py
PyExplorer: Yes, sure, let's move!

PyExplorer:
Thanks a mil @kmike for the final detailed reviews and valuable suggestions. Agree, that using mock approach in such projects is probably not the best choice, and I will definitely include rewriting tests for this pipeline here in my plans.

wRAR (approved):

Issue #89, opened 2024-04-07
split some modules to sub-packages


kmike:
It looks like some of the files grew too much, e.g. items.py, or even components.py. I think it makes sense to split them.

PR #92, opened 2024-04-13
Re-organize code

kmike:
This fixes GH-89.

Current approach:

1. There is a Python module per item type in zyte_common_items.items. Each module contains item, sub-item (if applicable) and metadata class.
2. Components are split into files. Related reusable components are put into modules, components specific to data types are put to separate files (e.g. components.real_estate).
3. New converters module is added, with attrs converters; RequestCaster should be deprecated, and MetadataCaster is not used directly anymore
4. request and metadata processors are moved to processors

TODO:

* [x] ~~review top-level imports (do we need all of them?)~~ (can be done separately)
* [x] split pages.py
* [x] ~~split processors.py~~ (can be done separately)
* [x] figure out a better way to organize MetadataCaster & friends
* [x] figure out a better way to organize RequestCaster & friends
* [ ] review API docs
* [x] ~~split test_items.py~~ (can be done separately)
* [x] fix mypy
* [x] remove unneeded future imports
* [x] Deprecations
  * [x] ~~components.cast_request~~ (not sure we need a deprecation)
  * [x] components.request_list_processor
  * [x] items.RequestListCaster
  * [x] ~~util.cast_metadata~~ (not sure we need a deprecation)
  * [x] util.metadata_processor
  * [x] util.MetadataCaster


zyte_common_items/components/metadata.py
```diff
@@ -0,0 +1,74 @@
...
+    #: requested, and that URL points to the webpage of a product with complete
+    #: certainty, the value should be `1`. If with complete certainty the
+    #: webpage features a job listing instead of a product, the value should be
+    #: `0`. When there is no complete certainty, the value could be anything in
+    #: between (e.g. `0.96`).
+    probability: Optional[float] = 1.0
+
+
+@attrs.define(kw_only=True)
+class ListMetadata(Item):
```
kmike, line 27: `_ListMetadata` and `_DetailMetadata` are made public, with updated docstrings.

zyte_common_items/components/business_place.py
```diff
@@ -0,0 +1,52 @@
...
+
+    #: Name of the parent place.
+    name: str
+
+    #: Identifier of the parent place.
+    placeId: str
+
+
+@attrs.define(kw_only=True)
+class OpeningHoursItem(Item):
```
kmike, line 33: I think we should rename it to OpeningHours

zyte_common_items/pages/search_request_template.py
```diff
@@ -0,0 +1,11 @@
...
+
+from zyte_common_items.items import SearchRequestTemplate, SearchRequestTemplateMetadata
+
+from .mixins import HasMetadata
+
+
+class SearchRequestTemplatePage(
+    ItemPage[SearchRequestTemplate], HasMetadata[SearchRequestTemplateMetadata]
+):
+    pass
```
kmike, line 11: This module misses Base and Auto page objects

zyte_common_items/components/request.py
```diff
@@ -0,0 +1,88 @@
...
+            **kwargs,
+        )
+
+    def cast(self, cls: Type[RequestT]) -> RequestT:
+        """Convert *value*, an instance of :class:`Request` or a subclass, into
+        *cls*, a different class that is also either :class:`Request` or a
+        subclass."""
+        new_value = convert_to_class(self, cls)
+        if type(self) is Request and cls is ProbabilityRequest:
+            new_value.metadata = ProbabilityMetadata(probability=1.0)  # type: ignore[attr-defined]
```
kmike, line 79: not sure why mypy warns about a missing attribute here

zyte_common_items/components/business_place.py
kmike: Actually, the current name is correct! It's a list item.

zyte_common_items/converters.py
```diff
@@ -0,0 +1,73 @@
...
+    :class:`~web_poet.ResponseUrl` object as a string."""
+
+    if not isinstance(url, (str, _Url)):
+        raise ValueError(
+            f"{url!r} is neither a string nor an instance of RequestUrl or ResponseUrl."
+        )
+    return str(url)
+
+
+def url_to_str_optional(url: Union[str, _Url, None]) -> Optional[str]:
```
kmike, line 22: It's not

```py
url_to_str_optional = attrs.converters.optional(url_to_str)
```

because mypy attrs plugin is stupid, and shows errors in this case, which require more type:ignores. The way it's implemented mypy understands it.

tests/test_pages.py
```diff
@@ -252,6 +243,8 @@ def allow_field(field_name):
         for prefix in ["_", "from_", "get_"]:
             if field_name.startswith(prefix):
                 return False
+        if field_name in {"cast"}:
+            return False
         return True
```
Gallaecio, line 248:
```suggestion
        if field_name == "cast":
            return False
        return True
```

kmike: I've added `{}` to handle a case if we add more methods in future, but it's probably pointless. Also, maybe all this dir-based code can be replaced with something from attrs.

zyte_common_items/components/metadata.py
```diff
@@ -0,0 +1,83 @@
...
+from zyte_common_items._dateutils import parse_iso_datetime
+from zyte_common_items.util import convert_to_class
+
+MetadataT = TypeVar("MetadataT", bound="BaseMetadata")
+
+
+class BaseMetadata(Item):
+    def cast(self, cls: Type[MetadataT]) -> MetadataT:
+        """Convert *value*, a metadata instance, into a different metadata *cls*."""
+        return convert_to_class(self, cls)
```
kmike, line 16: I've moved it to a method because it was not clear where to put the function otherwise; this casting is not a converter, and not a processor, and it's specific to metadata.

The disadvantage (or, maybe, a difference), as compared to the previous aproach, is that previously to convert something the first argument didn't have to be a metadata subclass, i.e. other classes, outside of our hierarchy, were supported. But I haven't came up with a use case for this.

zyte_common_items/components/request.py
```diff
@@ -0,0 +1,88 @@
...
+            method=self.method or "GET",
+            headers=header_list,
+            body=self.body_bytes,
+            **kwargs,
+        )
+
+    def cast(self, cls: Type[RequestT]) -> RequestT:
+        """Convert *value*, an instance of :class:`Request` or a subclass, into
+        *cls*, a different class that is also either :class:`Request` or a
+        subclass."""
```
kmike, line 76: https://github.com/zytedata/zyte-common-items/pull/92/files#r1565582009 applies here as well

zyte_common_items/pages/search_request_template.py
Gallaecio: On purpose. Should we provide them?

Gallaecio (commented):
:100:

zyte_common_items/util.py
```diff
@@ -48,38 +49,10 @@ def split_dict(dict: Dict, key_pred: Callable[[Any], Any]) -> Tuple[Dict, Dict]:
...
-
-
-def format_datetime(dt) -> str:
-    """Return the specified :class:`~datetime.datetime` object, assumed to be
-    in the UTC timezone, in ISO format, with the timezone specified as
-    ``Z``."""
-    return f"{dt.replace(tzinfo=None).isoformat(timespec='seconds')}Z"
-
-
-def parse_iso_datetime(date_str) -> datetime.datetime:
```
kmike, line 69: we don't need a backwards compat import for this, because it hasn't been released

Gallaecio (approved):

PR #90, opened 2024-04-10
cleanup CI

kmike:
This fixes most(?) warnings in tests.

PR #93, opened 2024-04-23
Release notes for 0.19.0.

wRAR:


CHANGELOG.rst
```diff
@@ -2,6 +2,22 @@
...
+===================
+
+* Now requires ``attrs >= 22.2.0``.
+
+* Added the :class:`~.DropLowProbabilityItemPipeline` that drops items with the
+  ``probability`` value lower than a set threshold.
+
+* Added the :attr:`~.ListMetadata.validationMessages` attribute.
+
+* Reorganized the module structure.
```
wRAR, line 15: Should this be more verbose?

Gallaecio: Since most (all?) changes are internal, this is probably OK.

CHANGELOG.rst
```diff
@@ -2,6 +2,22 @@
...
+* Added :class:`~.pipelines.DropLowProbabilityItemPipeline` that drops items
+  with the ``probability`` value lower than a set threshold.
+
+* Added the :attr:`.ListMetadata.validationMessages` attribute.
+
+* Reorganized the module structure.
+
+* Documentation improvements.
+
+* Test and CI fixes and improvements.
```
kmike, line 19: There is a sneaky change in https://github.com/zytedata/zyte-common-items/pull/90/: new ListMetadata.get_date_downloaded_parsed method

CHANGELOG.rst
kmike: There are also some classes which are promoted to public, e.g. BaseMetadata, ListMetadata, DetailsMetadata

kmike: We may also mention zyte_common_items.converters module; the code was scattered in various places before, but now it contains useful attrs convertors in a single place

CHANGELOG.rst
wRAR: I've noticed it, but wasn't sure if it's intended to be used outside tests.

docs/reference/converters.rst
```diff
@@ -0,0 +1,6 @@
+==========
+Converters
+==========
+
+.. automodule:: zyte_common_items.converters
```
wRAR, line 5: Getting `WARNING: py:class reference target not found: web_poet.page_inputs.url._Url` here, but as it's indeed not mentioned in the web-poet docs I guess there is nothing to do?

CHANGELOG.rst
kmike: Just checked it; it seems that's only these 3 classes :)

docs/reference/converters.rst
kmike: I think that's ok, if it doesn't break CI

PR #95, opened 2024-06-17
Implement auto_field field metadata

Gallaecio:
Suggested by @kmike [here](https://github.com/scrapy-plugins/scrapy-zyte-api/pull/202#discussion_r1642855504).

### DIFF
--- a/docs/usage/pages.rst
+++ b/docs/usage/pages.rst
@@ -45,6 +45,32 @@ whose ``to_item`` method returns an instance of
         def name(self):
             return self.css("h1::text").get()
 
+.. _extractors:
+
+Extractors
+==========
+
+For some nested fields (:class:`~.ProductFromList`, :class:`~.ProductVariant`),
+:ref:`base extractors <default-processors-nested>` exist that you can subclass
+to write your own extractors.
+
+They provide the following base line:
+
+-   They declare the :ref:`item class <items>` that they return, allowing for
+    their ``to_item`` method to automatically build an instance of it from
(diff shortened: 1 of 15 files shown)


docs/usage/pages.rst
```diff
@@ -72,23 +98,24 @@ extra fields. For example:
...
+    print(is_auto_field(ExtendedProductPage, "name"))  # Returns False
+    print(is_auto_field(ExtendedProductPage, "foo"))  # Returns False
+    print(is_auto_field(ExtendedProductPage, "brand"))  # Returns True
+    print(is_auto_field(ExtendedProductPage, "bar"))  # Raises KeyError
 
-See :ref:`extractor-api`.
+If you are overriding a field method but the method continues to return the
+value straight from the ``Auto``-prefixed class, e.g. because you are only
+overriding field metadata or :ref:`processors <processors>`, you should also
+set ``auto_field`` to ``True``. Instead of setting it manually in the field
```
kmike, line 117: I think it depends on how we want to use it. For example, if an item value comes from Zyte API, but a processor is needed, I can imagine the value was not right. In this case, an override is needed, and probably it's better to indicate it by not using auto_field.

kmike: I'm not 100% sure what are the scenarios where you do want to mark it as auto (maybe some stats, logging, etc?).

kmike: We also have an option to not document the scenarios for now.

PR #96, opened 2024-06-18
Release notes for 0.20.0

Gallaecio:
~It’s not really a bug fix, so maybe 0.20.0 would make sense, but it feels quite insignificant, so I’m not sure.~

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,7 +2,7 @@
 Changelog
 =========
 
-0.20.0 (2024-06-DD)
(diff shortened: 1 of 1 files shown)


kmike:
I think 0.20 makes sense :) It's insignificant, but it's a new feature, new imports available, etc. If a package needs this new feature, it'd be a bit weird it needs zyte-common-items 0.19.1, and doesn't work with 0.19.0.

wRAR (approved):

PR #97, opened 2024-06-20
add 'notes' to Request

BurnzZ:
A proposal to add some way of putting some kind of information to the Request.

One use case is adding some identifiers to a given request. For example, in https://github.com/zytedata/zyte-spider-templates/blob/main/zyte_spider_templates/pages/product_navigation_heuristics.py#L72, a current workaround would be to add some kind of _url prefix_ to `request.name`.

It's not pretty but it does the job. However, it'd be hard to properly identify some information here.

It'd be nice if we have some dedicated field to put such arbitrary values. This would be quite useful when developers create page objects with some added notes to the `Request`s.

### DIFF
--- a/tests/test_components.py
+++ b/tests/test_components.py
@@ -143,12 +143,12 @@ def test_probability_request():
     request = ProbabilityRequest(
         name="Get with RequestURL object",
         url=RequestUrl("https://example.com/test"),
-        notes=["arbitrary", "info"],
+        notes={"type": "test request", "id": 782},
     )
 
     assert request.url == "https://example.com/test"
     assert request.get_probability() is None
-    assert request.notes == ["arbitrary", "info"]
+    assert request.notes == {"type": "test request", "id": 782}
 
 
 def test_request():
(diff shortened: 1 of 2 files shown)


Gallaecio:
Something like `Request.meta: Dict[str, Any]`? :slightly_smiling_face:

BurnzZ:
> Something like `Request.meta: Dict[str, Any]`? 🙂

I initially implemented it using a `dict` but quickly realized I don't have something to put in as values. 😅 However, I do see the point in having `dict` here since:

- Faster to find something because of the keys.
- More convenient to structure the data. For example, `{"id": 312, "type": "heuristic", "type": "productNavigation"}`

Updated this in https://github.com/zytedata/zyte-common-items/pull/97/commits/a92ca86a5e635518e41cc9fda6607f3ed5c9d998.

> I wonder if, to make it be in line with the rest of the schema, it should be called metadata or additionalMetadata, and be in the same format as additionalAttributes.

Not quite sure yet with this one. I think I'm slightly leaning towards it not having to conform with the rest of the schema just to give the idea that values from this field doesn't come from Zyte API Extraction. I was thinking of calling it `user_notes` but that's too long.

wRAR (approved):

wRAR:
I like the dict type better, I'm not sure if I like the "notes" name but I don't think "*metadata" is better.

kmike:
+1 to using dicts. 

I think it looks more like metadata, not "notes", so +1 to something along these lines.

ProbabilityRequest already have metadata attribute though. Having both "meta" and "metadata" sounds weird. For me it feels more natural to put metadata into the existing metadata attribute.

But how to do it? I think we can consider extending BaseMetadata to include this dict; it'd be something like `request.metadata.non_standard` or `request.metadata.custom`. This is a big change, because this "non_standard" metadata is going to be available for all other item types as well.

PR #98, opened 2024-08-01
Add Serp

Gallaecio:


PyExplorer (approved):

zyte_common_items/pages/serp.py
```diff
@@ -0,0 +1,40 @@
...
+
+from .base import BasePage, Page
+from .mixins import HasMetadata
+
+
+class BaseSerpPage(BasePage, Returns[Serp], HasMetadata[SerpMetadata]):
+    pass
+
+
+class SerpPage(Page, Returns[Serp], HasMetadata[SerpMetadata]):
```
PyExplorer, line 18: Just wondering, how and where will `SerpPage` be used?

Gallaecio: It should work as a base class for custom page objects for Serp items.

BurnzZ (approved):
LGTM!

VMRuiz:
The fields that defined so far looks good. However, some other important fields like subtitlet, or media type - videos, images - are missing. Are you implementing only the essential fields at this point and leaving the rest of feature revisions?

zyte_common_items/components/__init__.py
```diff
@@ -21,4 +21,5 @@
 from .ratings import AggregateRating, StarRating
 from .real_estate import RealEstateArea
 from .request import Header, ProbabilityRequest, Request
+from .serp import SerpOrganicResult
```
kmike, line 24: Currently for other listing items list elements are not in the components module, they're in items directly - see e.g. product list or article list.

Issue #94, opened 2024-05-17
Standard price processors


bulatbulat48:
Standard price processors at the from ProductPage don't process integer and float values.

description:
Supported inputs are [Selector](https://parsel.readthedocs.io/en/stable/parsel.html#parsel.selector.Selector), [SelectorList](https://parsel.readthedocs.io/en/stable/parsel.html#parsel.selector.SelectorList) and [HtmlElement](https://lxml.de/apidoc/lxml.html.html#lxml.html.HtmlElement). Other inputs are returned as is.

https://zyte-common-items.readthedocs.io/en/latest/reference/field-processors.html#zyte_common_items.processors.price_processor

nodes decorator:
https://github.com/zytedata/zyte-common-items/blob/main/zyte_common_items/processors.py#L118

PR #99, opened 2024-08-12
Processor changes

Nykakin:
We'd like to remove processors from our utils and replace them with `zyte-common-items` counterparts. There are some functionality missing, however.

Summary of changes:
* Add new processor, `images_processor`, that attempts to convert input to list of `zyte_common_items.Image` instances
* Make `brand_processor` return `zyte_common_items.Brand` instances instead of strings
* Allow `brand_processor` to take string as an input argument
* Allow `price_processor` to take ~~string or~~ number as input argument (resolves https://github.com/zytedata/zyte-common-items/issues/94)

### DIFF
--- a/tests/test_pages_price.py
+++ b/tests/test_pages_price.py
@@ -62,21 +62,21 @@ async def test_price_explicit():
     url = "https://example.com"
     page = CustomProductPage(response=HttpResponse(url=url, body=html))
     assert page.call_count == 0
-    assert page.price == "13.20"
+    assert page.price == "$13.2"
     assert page.call_count == 1
     assert page.currency is None
-    assert await page.currencyRaw == "$"
-    assert page.call_count == 1  # we want this to be 1
-    assert await page.currencyRaw == "$"
-    assert page.call_count == 1  # we want this to be 1
+    assert await page.currencyRaw is None
+    assert page.call_count == 2  # we want this to be 1
+    assert await page.currencyRaw is None
+    assert page.call_count == 2  # we want this to be 1
 
     # access currency fields before the price field
     page = CustomProductPage(response=HttpResponse(url=url, body=html))
     assert page.call_count == 0
     assert page.currency is None
-    assert await page.currencyRaw == "$"
+    assert await page.currencyRaw is None
     assert page.call_count == 1
-    assert page.price == "13.20"
+    assert page.price == "$13.2"
     assert page.call_count == 2  # we want this to be 1
 
 
--- a/tests/test_pages_price.py
+++ b/tests/test_pages_price.py
@@ -62,21 +62,21 @@ async def test_price_explicit():
     url = "https://example.com"
     page = CustomProductPage(response=HttpResponse(url=url, body=html))
     assert page.call_count == 0
-    assert page.price == "$13.2"
+    assert page.price == "13.20"
     assert page.call_count == 1
     assert page.currency is None
-    assert await page.currencyRaw is None
-    assert page.call_count == 2  # we want this to be 1
-    assert await page.currencyRaw is None
-    assert page.call_count == 2  # we want this to be 1
+    assert await page.currencyRaw == "$"
+    assert page.call_count == 1  # we want this to be 1
+    assert await page.currencyRaw == "$"
+    assert page.call_count == 1  # we want this to be 1
 
     # access currency fields before the price field
     page = CustomProductPage(response=HttpResponse(url=url, body=html))
     assert page.call_count == 0
     assert page.currency is None
-    assert await page.currencyRaw is None
+    assert await page.currencyRaw == "$"
     assert page.call_count == 1
-    assert page.price == "$13.2"
+    assert page.price == "13.20"
     assert page.call_count == 2  # we want this to be 1
 
 
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -136,7 +136,9 @@ def test_breadcrumbs_base_url():
     [
         (None, None),
         ("", None),
+        (" ", None),
         ("foo", Brand(name="foo")),
+        (" foo ", Brand(name="foo")),
         (Selector(text="<html></html>"), None),
         (SelectorList([]), None),
         (fromstring("<p>foo</p>"), Brand(name="foo")),
@@ -401,6 +403,8 @@ def test_images_page():
         ({}, {}),
         (22.9, "22.90"),
         (22.0, "22.00"),
+        ("22.9", "22.9"),
+        ("Do not apply to strings...", "Do not apply to strings..."),
     ],
 )
 def test_prices(input_value, expected_value):
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -15,7 +15,6 @@ from zyte_common_items import (
     Gtin,
     Image,
     ProductPage,
-    ProductVariant,
 )
 from zyte_common_items.processors import (
     _format_price,
@@ -25,7 +24,6 @@ from zyte_common_items.processors import (
     images_processor,
     price_processor,
     rating_processor,
-    variants_processor
 )
 
 base_url = "http://www.example.com/blog/"
@@ -406,22 +404,3 @@ def test_prices(input_value, expected_value):
 
     page = PricePage(base_url)  # type: ignore[arg-type]
     assert page.price == expected_value
-
-
-@pytest.mark.parametrize(
-    "input_value,expected_value",
-    [
-        (
-            [{"price": "12,12$"}, {"price": 100}],
-            [ProductVariant(price="12.12"), ProductVariant(price="100.00")]
-        ),
-    ],
-)
-def test_variants(input_value, expected_value):
-    class VariantPage(BasePage):
-        @field(out=[variants_processor])
-        def variants(self):
-            return input_value
-
-    page = VariantPage(base_url)  # type: ignore[arg-type]
-    assert page.variants == expected_value
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -382,17 +382,10 @@ def test_images(input_value, expected_value):
 @pytest.mark.parametrize(
     "input_value,expected_value",
     [
-        ("$10", "10.00"),
-        ("100  ", "100.00"),
-        ("100rub", "100.00"),
         (100, "100.00"),
         (None, None),
         ([], []),
         ({}, {}),
-        ("", None),
-        ("buy 10 ab ab", "10.00"),
-        ("1,000.17", "1000.17"),
-        ("1,000", "1000.00"),
         (22.9, "22.90"),
         (22.0, "22.00"),
     ],
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -14,7 +14,6 @@ from zyte_common_items import (
     Breadcrumb,
     Gtin,
     Image,
-    Page,
     ProductPage,
 )
 from zyte_common_items.processors import (
@@ -159,8 +158,8 @@ def test_brand(input_value, expected_value):
 
 
 def test_brand_page():
-    class MyProductPage(Page):
-        @field(out=[brand_processor])
+    class MyProductPage(ProductPage):
+        @field
         def brand(self):
             return self.css("body")
 
@@ -377,6 +376,20 @@ def test_images(input_value, expected_value):
 
     page = ImagesPage(base_url)  # type: ignore[arg-type]
     assert page.images == expected_value
+
+
+def test_images_page():
+    class MyProductPage(ProductPage):
+        @field
+        def images(self):
+            return self.css("img::attr(href)").getall()
+
+    response = HttpResponse(
+        url="http://www.example.com/",
+        body="<html><body><img href='https://www.url.com/img.jpg'></body></html>".encode(),
+    )
+    page = MyProductPage(response=response)
+    assert page.images == [Image(url="https://www.url.com/img.jpg")]
 
 
 @pytest.mark.parametrize(
--- a/tests/test_processors.py
+++ b/tests/test_processors.py
@@ -14,6 +14,7 @@ from zyte_common_items import (
     Breadcrumb,
     Gtin,
     Image,
+    Page,
     ProductPage,
 )
 from zyte_common_items.processors import (
@@ -158,7 +159,7 @@ def test_brand(input_value, expected_value):
 
 
 def test_brand_page():
-    class MyProductPage(BasePage):
+    class MyProductPage(Page):
         @field(out=[brand_processor])
         def brand(self):
             return self.css("body")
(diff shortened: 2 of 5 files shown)


tests/test_pages_price.py
```diff
@@ -62,21 +62,21 @@ def price(self):
...
-    assert page.price == "$13.2"
+    assert page.price == "13.20"
     assert page.call_count == 1
     assert page.currency is None
-    assert await page.currencyRaw is None
-    assert page.call_count == 2  # we want this to be 1
-    assert await page.currencyRaw is None
-    assert page.call_count == 2  # we want this to be 1
+    assert await page.currencyRaw == "$"
+    assert page.call_count == 1  # we want this to be 1
```
wRAR, line 69: I wonder why did the count change and what should be the new expectation in the comment.

zyte_common_items/processors.py
```diff
@@ -104,50 +107,80 @@ def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
     return results
 
 
-@only_handle_nodes
-def brand_processor(value: Union[Selector, HtmlElement], page: Any) -> Any:
+def brand_processor(value: Any, page: Any) -> Union[Brand, None]:
```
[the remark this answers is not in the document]
Nykakin: Changed in ba1a5e93749c65eac05e4879588f87695aca818e

zyte_common_items/processors.py
```diff
@@ -104,50 +107,80 @@ def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
...
-    Supported inputs are :class:`~parsel.selector.Selector`,
-    :class:`~parsel.selector.SelectorList` and :class:`~lxml.html.HtmlElement`.
-    Other inputs are returned as is.
+    If inputs are either :class:`~parsel.selector.Selector`,
+    :class:`~parsel.selector.SelectorList` or :class:`~lxml.html.HtmlElement`, attempts
+    to extract brand data from it.
+
+    If value is a string, use it to create brand object instance
+
+    Other inputs are returned unchanged
```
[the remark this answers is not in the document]
Nykakin: Changed in Changed in ba1a5e93749c65eac05e4879588f87695aca818e

tests/test_pages_price.py
Nykakin: Yes, I've noticed that comment info doesn't match test values in places in master. Not sure if this is a proper behavior.

On master such difference can be found in 5 places:
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_pages_price.py#L41 
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_pages_price.py#L69
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_pages_price.py#L71
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_pages_price.py#L80
* https://github.com/zytedata/zyte-common-items/blob/main/tests/test_pages_price.py#L188

This PR fixes two of these:
* https://github.com/zytedata/zyte-common-items/blob/6620a1538ee3b558d78705800ad4ea19f058f7dc/tests/test_pages_price.py#L69
* https://github.com/zytedata/zyte-common-items/blob/6620a1538ee3b558d78705800ad4ea19f058f7dc/tests/test_pages_price.py#L71
While three are same as before:
* https://github.com/zytedata/zyte-common-items/blob/6620a1538ee3b558d78705800ad4ea19f058f7dc/tests/test_pages_price.py#L41
* https://github.com/zytedata/zyte-common-items/blob/6620a1538ee3b558d78705800ad4ea19f058f7dc/tests/test_pages_price.py#L80
* https://github.com/zytedata/zyte-common-items/blob/6620a1538ee3b558d78705800ad4ea19f058f7dc/tests/test_pages_price.py#L188

It seems like an improvement, albeit incomplete (but fixing this for all cases seems kind of out of scope so perhaps we can live with it for now...?)

zyte_common_items/processors.py
```diff
@@ -330,6 +363,45 @@ def aggregateRating(self):
     return value
 
 
+def images_processor(value: Any, page: Any) -> Any:
```
kmike, line 366: it seems page argument is unused - can we remove it?

zyte_common_items/processors.py
```diff
@@ -330,6 +363,45 @@ def aggregateRating(self):
...
+
+    If the input is a string, it's used as a url for returning image object.
+
+    If input is either an iterable of strings or mappings with "url" key, they are
+    used to populate image objects.
+
+    Other inputs are returned unchanged.
+    """
+
+    value = _handle_selectorlist(value)
```
kmike, line 378: Why is this needed? It seems that if a SelectorList is passed, the first item is taken, and then returned as-is. But it doesn't seem to be really supported, as in the end it's not a valid Image instance (as you have in TODO). So, it seems for now it can be dropped.

zyte_common_items/processors.py
kmike: Although I see other processors are written this way as well; it makes sense to keep it for consistency for now, and address later if needed.

zyte_common_items/processors.py
```diff
@@ -104,50 +107,80 @@ def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
...
     Other inputs are returned as is.
 
     .. _price-parser: https://github.com/scrapinghub/price-parser
     """
-    price = extract_price(value)
-    return _format_price(price)
+    value = _handle_selectorlist(value)
+
+    if isinstance(value, Real):
+        return f"{value:.2f}"
```
kmike, line 178: What do you think about moving it to https://github.com/zytedata/zyte-parsers/blob/main/zyte_parsers/price.py?

kmike: It seems this would allow to remove the duplication in price_processor vs simple_price_processor.

tests/test_processors.py
```diff
@@ -149,7 +159,7 @@ def brand(self):
 
 def test_brand_page():
     class MyProductPage(ProductPage):
-        @field
+        @field(out=[brand_processor])
```
kmike, line 162: why is this needed? Shouldn't subclasses of ProductPage have brand_processor enabled by default?

tests/test_processors.py
```diff
@@ -321,3 +331,76 @@ def aggregateRating(self):
...
+            [
+                Image("https://www.url.com/img1.jpg"),
+                Image("https://www.url.com/img2.jpg"),
+            ],
+        ),
+    ],
+)
+def test_images(input_value, expected_value):
+    class ImagesPage(BasePage):
+        @field(out=[images_processor])
```
kmike, line 373: Does it work without `out`, as images_processor is enabled by default in this PR? It'd be nice to test it without `out`, because it's going to be the most important use case (default behavior).

tests/test_processors.py
```diff
@@ -321,3 +331,76 @@ def aggregateRating(self):
...
+        ("buy 10 ab ab", "10.00"),
+        ("1,000.17", "1000.17"),
+        ("1,000", "1000.00"),
+        (22.9, "22.90"),
+        (22.0, "22.00"),
+    ],
+)
+def test_prices(input_value, expected_value):
+    class PricePage(BasePage):
+        @field(out=[price_processor])
```
kmike, line 401: is `out` needed?

tests/test_pages_price.py
```diff
@@ -62,21 +62,21 @@ def price(self):
     url = "https://example.com"
     page = CustomProductPage(response=HttpResponse(url=url, body=html))
     assert page.call_count == 0
-    assert page.price == "$13.2"
+    assert page.price == "13.20"
```
kmike, line 65: This is a big change, because it goes against the documentation - see https://zyte-common-items.readthedocs.io/en/latest/usage/field-processors.html#overview:

> By design, the processors enabled by default are “transparent”: they don’t change the output of the field if the result is of the expected final type. For example, if there is a [str](https://docs.python.org/3/library/stdtypes.html#str) attribute in the item, and the field returns [str](https://docs.python.org/3/library/stdtypes.html#str) value, the default processor returns the value as-is.

This design decision was made for two reasons:

1. It makes it possible to safely add new default processors, without a risk of breaking user code. If the code was working before a processor is introduced, it means the field was returning the final data type. Processor won't touch it, so no breaking is possible.
2. it provides an escape hatch for the user; if you want to return a final data, you can do it just by returning the value.

It's not an issue for Real support in price parsing, for Brand processing, or for Images (it all looks good in this PR), but it's an issue for price fields.

Obviously, this "design principle" has a downside: in most cases, we do want to process strings by default. I think most large Scrapy projects develop this kind of utilities over time, where all values are processed by default. We do want to make it more standard.

I can see 2 ways to address it. 

1. Start procesing final data types by default; drop this design principle.
2. Introduce more aggressive / complete versions of procesors. Keep "cautious" processors enabled by default, which don't handle the final data types. Make it easy to enable the complete versions - either by defining Processors class which users can use, or by providing another base ProductPage class, with processing of final values enabled by default.

I'm +0.5 for option (2), but willing to hear the opinions and arguments. I like keeping the "safe" approach, but it introduces more complexity (e.g. an additional base class, where users need to pick the right one). 

cc @wRAR @Gallaecio.

kmike:
Overall, 👍, much needed changes. Having e.g. an images processor is a must.

zyte_common_items/processors.py
Nykakin: Moved it inside `TODO` comment in d2ab672afffdc9aad5a064a8345e663d4af8dd33 since indeed it isn't needed for now.

zyte_common_items/processors.py
Nykakin: We'd like to migrate our stuff quickly before certain deadlines, which is why I want to have this thing working for now and avoid coordinating between PR in yet another dependency repository (which is also why I placed proposed parsing of images into TODO). I think we can improve it later and migrate into more fitting place.

tests/test_processors.py
Nykakin: Yes, replaced `ProductPage` with `BasePage` in d2ab672afffdc9aad5a064a8345e663d4af8dd33.

tests/test_processors.py
Nykakin: These are tests for processors, not particular pages, which is why I think it makes sense to limit to `BasePage`. Copied this approach from other test in this file, such as [test for Gtin processor](https://github.com/zytedata/zyte-common-items/blob/main/tests/test_processors.py#L221-L228). After all, same processor can be applied to many different page types.

tests/test_processors.py
Nykakin: In this case yes, base class is `BasePage` so there's no default processor for `price` field and I need to mark it explicitly.

tests/test_pages_price.py
Nykakin: I'm kind of against this transparency principle, and I've even proposed another PR https://github.com/zytedata/zyte-common-items/pull/101 to ensure all strings are trimmed.

It's difficult to imagine for me why we would want to emit data with extra whitespaces and it's bothersome to always manually add some sort of default text cleaning processor into all used fields. I think this kind of processing should be applied by default at some point and would help to ensure all our output data respect a common standard.

tests/test_processors.py
Nykakin: Actually it needs to be `Page`, not `BasePage`, because `self.css` is not working otherwise, updated in 30baf8a0ca371d93af1118a7829d8f4e607378a8

Nykakin:
In response to https://github.com/zytedata/zyte-common-items/pull/99#discussion_r1725091519 and after some discussion in slack I've decided to just return strings without any change in `price_processor` instead of parsing it, thus retaining original philosophy of transparent processors.

The discussion about this approach can continue but I do not want to block image processors and brand processor because of it.

tests/test_processors.py
Nykakin: Looks like code coverage tool is complaining, so in the end I had to remove `brand_processor` as originally suggested. Fixed in aa42e22b7db52fc5ab1a711df44f6180069d6d37

kmike (approved):
👍

zyte_common_items/processors.py
kmike: Sounds good!

tests/test_processors.py
```diff
@@ -321,3 +331,83 @@ def aggregateRating(self):
...
+
+@pytest.mark.parametrize(
+    "input_value,expected_value",
+    [
+        (100, "100.00"),
+        (None, None),
+        ([], []),
+        ({}, {}),
+        (22.9, "22.90"),
+        (22.0, "22.00"),
```
Gallaecio, line 403: Should we add a test scenario that confirms that strings are left as is, e.g. `("22.9", "22.9")`?

zyte_common_items/processors.py
```diff
@@ -104,50 +107,78 @@ def _from_zp_breadcrumb(value: zp_Breadcrumb) -> Breadcrumb:
...
+
+    If value is a string, uses it to create a :class:`~zyte_common_items.Brand` instance.
+
+    Other inputs are returned unchanged.
     """
-    return extract_brand_name(value, search_depth=2)
+    value = _handle_selectorlist(value)
 
+    if isinstance(value, str):
+        return Brand(name=value) if value else None
```
Gallaecio, line 124: @kmike Do you think stripping would make sense here? Or if we want to keep the original spacing for non-empty strings, at least do `… if value.strip() else …`?

zyte_common_items/processors.py
```diff
@@ -330,6 +361,44 @@ def aggregateRating(self):
...
+
+    Other inputs are returned unchanged.
+    """
+
+    # TODO: add generic-purpose extract_images utility to zyte-parsers
+    #
+    # value = _handle_selectorlist(value)
+    # if isinstance(value, (Selector, HtmlElement)):
+    #    images = extract_images(value)
+    #    return [Image(url=url) for url in images]
```
Gallaecio, line 381: Could we move this comment into a repository issue instead?

Gallaecio (approved):

zyte_common_items/processors.py
kmike: I think stripping is fine here.

zyte_common_items/processors.py
Nykakin: Removed comment in 12cb8efadd59d399b41bf59de6b0d721d4d10efb created issue https://github.com/zytedata/zyte-parsers/issues/16

zyte_common_items/processors.py
Nykakin: Added `strip()` and some more test cases in 12cb8efadd59d399b41bf59de6b0d721d4d10efb

tests/test_processors.py
Nykakin: Added new test cases in 12cb8efadd59d399b41bf59de6b0d721d4d10efb

kmike:
Thanks @Nykakin!

PR #100, opened 2024-08-15
Add the customAttributes field to items and pages.

wRAR:


wRAR:
Not sure if this should go into a base class (e.g. `Item`) as we currently don't have any fields in it.

Gallaecio (approved):

Gallaecio:
I do wonder if we should consider deprecating `_unknown_fields_dict` and instead mapping those into `customAttributes`. Although things could be problematic if an item has both, e.g. `{"foo": "bar", "customAttributes": {"foo": "baz"}}`.

PR #101, opened 2024-08-21
Always trim string fields

Nykakin:
Always trim string fields.

For all fields of type `Optional[str]` or `str` apply `str.strip()` through processors.

For all fields of type `Optional[List[str]]` or `List[str]`  apply `str.strip()` to all elements through processors.

Extract processors to separate classes following approach found in [extractors](https://github.com/zytedata/zyte-common-items/blob/main/zyte_common_items/extractors.py#L7-L9) to reduce duplication in code.


zyte_common_items/processors.py
```diff
@@ -345,3 +345,21 @@ def metadata_processor(metadata: BaseMetadata, page):
     if page.metadata_cls is None:
         return None
     return metadata.cast(page.metadata_cls)
+
+
+def string_processor(value: str) -> Union[str, None]:
+    """Processor for string values"""
+    if isinstance(value, str):
+        return value.strip()
+    return value
```
Nykakin, line 354: Perhaps we should return `None` if value is not a string instance...?

tests/test_processors.py
```diff
@@ -321,3 +323,43 @@ def aggregateRating(self):
...
+
+
+@pytest.mark.parametrize(
+    "input_value,expected_value",
+    [
+        (None, None),
+        ("", ""),
+        ("Value ", "Value"),
+        (" Value", "Value"),
+        (" Value ", "Value"),
```
Gallaecio, line 335: `("foo bar", "foo bar"),` (i.e. make sure we do not remove spaces between words)

zyte_common_items/processors.py
Gallaecio: I feel like we should convert it to `str` unless it is `None`.

I also wonder if we should return `None` if `not value.strip()`.

zyte_common_items/processors.py
```diff
@@ -345,3 +345,21 @@ def metadata_processor(metadata: BaseMetadata, page):
...
+
+
+def list_processor(processor: Callable) -> Any:
+    """Apply processor to a list of items"""
+
+    def loop(values):
+        if not isinstance(values, Iterable):
+            return values
+        return [processor(value) for value in values]
```
Gallaecio, line 363: I wonder if we should skip `None` values in the resulting list.

Gallaecio (commented):
+0.5 to these processors.

Even if we are processing `str`, I feel it is justified to meet the schema requirements, which I feel are for all these fields to be trimmed, even if [the schema only says that explicitly for article body](https://docs.zyte.com/zyte-data/schemas/article/1.0.html#articleBody).

PR #102, opened 2024-08-26
Prevent None.cast call

Gallaecio:


zyte_common_items/processors.py
```diff
@@ -405,6 +405,6 @@ def probability_request_list_processor(
 def metadata_processor(metadata: BaseMetadata, page):
```
kmike, line 405: should it be Optional[BaseMetadata] now?

kmike:
👍

PR #104, opened 2024-08-28
Handle numeric values with commas in rating processor

Nykakin:
Ratting processor currently accept dictionaries with standard keys `ratingValue`, `bestRating` and `reviewCount`. It attempts to convert values using `float` and `int` calls, so that string values are handled:
```python
>>> from zyte_common_items.processors import rating_processor
>>> 
>>> rating_processor({"bestRating": 5, "ratingValue": "3.3", "reviewCount": 12}, None)
AggregateRating(bestRating=5.0, ratingValue=3.3, reviewCount=12)
```
If the value of rating uses comma instead of a dot to separate decimal places, an exception is raised:
```python
>>> rating_processor({"bestRating": 5, "ratingValue": "3,3", "reviewCount": 12}, None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/tmp/zyte_common_items/processors.py", line 354, in rating_processor
    result.ratingValue = float(rating_value)
                         ^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: '3,3'
```
Similarly, number of reviews can have comma in string value (seen in Amazon pages), e.g. `12,070 ratings`. Strings such as these also raise exceptions.

Because standard schema defines that `ratingValue` and `bestRating` use `Number` type and `revewCount` uses `Integer` type, decided to treat strings differently depending on used type:
* If we expect real value, try replacing `,` with `.`
* If we expect integer value, try removing `,`

### DIFF
--- a/zyte_common_items/processors.py
+++ b/zyte_common_items/processors.py
@@ -304,7 +304,7 @@ def rating_processor(value: Any, page: Any) -> Any:
     The input can also be a dictionary with one or more of the
     :class:`~zyte_common_items.AggregateRating` fields as keys. The values for
     those keys can be either final values, to be assigned to the corresponding
-    fields, or selector-like objects.
+    fields, strings to be parsed, or selector-like objects.
 
     If a returning dictionary is missing the ``bestRating`` field and
     ``ratingValue`` is a selector-like object, ``bestRating`` may be extracted.
--- a/zyte_common_items/processors.py
+++ b/zyte_common_items/processors.py
@@ -1,5 +1,5 @@
 from collections.abc import Iterable, Mapping
-from functools import partial, wraps
+from functools import wraps
 from numbers import Real
 from typing import Any, Callable, List, Optional, Union, Type
 
@@ -30,17 +30,23 @@ from .components import (
     Request,
 )
 
-def _to_number(value: Any, number_type: Type) -> Any:
+def _to_int(value) -> Any:
     if isinstance(value, Real):
-        return number_type(value)
+        return int(value)
+    elif isinstance(value, str):
+        if "," in value:
+            value = value.replace(",", "")
+        return int(value)
+    return value
+
+def _to_float(value) -> Any:
+    if isinstance(value, Real):
+        return float(value)
     elif isinstance(value, str):
         if "," in value:
             value = value.replace(",", ".")
-        return number_type(value)
+        return float(value)
     return value
-
-_to_int = partial(_to_number, number_type=int)
-_to_float = partial(_to_number, number_type=float)
 
 
 def _get_base_url(page: Any) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


zyte_common_items/processors.py
```diff
@@ -31,6 +31,26 @@
...
+    if isinstance(value, Integral):
+        return int(value)
+    elif isinstance(value, str):
+        if "," in value:
+            value = value.replace(",", "")
+        return int(value)
+    return value
+
+
+def _to_float(value: Any) -> Any:
```
Gallaecio, line 44: I wonder if we should call it something different, like `to_rating_float`, to make it somewhat clear that it should not be used for values that could be greater than 999.

Gallaecio (approved):

zyte_common_items/processors.py
Nykakin: What about moving these as local functions within `rating_processor`? Then it would be clear their usage is restricted to this function.
```python
def rating_processor(value: Any, page: Any) -> Any:
    def _to_int(value: Any) -> Any:
        ...

    def _to_float(value: Any) -> Any:
        ...

    ...
```
That was my original idea but I've thought that perhaps we could re-use these in some other place so I left them at the top level.

wRAR: I think renaming is better as we can still reuse the renamed ones if we want.

Nykakin: @wRAR what naming would you propose...? Because `to_rating_float` sounds like something that should be used only for, well, rating.

Gallaecio: I agree that these functions could be useful elsewhere. I suggested `to_rating_float`, but something more generic, like `to_small_float`, would also work.

Nykakin: Wouldn't guess from name alone what a function `to_small_float` does.

Gallaecio: I’m open to other names.

But we don’t really need for the name to make it obvious what the function can do, I think the most important think is for the name to make people think it twice before they use the function for a value that could surpass 999. The details can be explained in a docstring.

kmike: FWIW, we also have https://github.com/zytedata/zyte-parsers/blob/8b161cf69df8caeb145e5244601ff93e41b287e6/zyte_parsers/aggregate_rating.py#L101 which does a similar thing for floats. I like how ints are handled in this PR though.

zyte_common_items/processors.py
```diff
@@ -31,6 +31,26 @@
...
+    if isinstance(value, Real):
+        return float(value)
+    elif isinstance(value, str):
+        if "," in value:
+            value = value.replace(",", ".")
```
kmike, line 49: nitpick: why is `if` necesary? Is it an optimization, and if so - does it actually help (it can go both ways - it's 2 passes over string instead of one)?

zyte_common_items/processors.py
kmike: The function in zyte-parsers also swallows the exceptions, and here it's not; it seems not silencing them is better for the case processor is processing the value returned by human-written code.

PR #111, opened 2024-11-06
Forum thread support

Gallaecio:


Gallaecio:
It’s probably just laziness on my end, but I wonder if it makes sense to keep adding class-specific tests like those in `test_components` and `test_items` for every new component and item. What do you think?

kmike:
I don't see a huge value in these tests.

kmike (approved):

Gallaecio:
(don’t ask why I did not include this in 0.25.0 :sweat_smile:)

PR #112, opened 2024-11-08
keyword → query

Gallaecio:


zyte_common_items/items/search_request_template.py
```diff
@@ -1,20 +1,33 @@
...
 import jinja2
+from jinja2 import StrictUndefined
+from jinja2.exceptions import UndefinedError
 
 from zyte_common_items.base import Item
 from zyte_common_items.components import DetailsMetadata, Header, Request
 from zyte_common_items.converters import to_metadata_optional
 
-_TEMPLATE_ENVIRONMENT = jinja2.Environment()
+_TEMPLATE_ENVIRONMENT = jinja2.Environment(undefined=StrictUndefined)
```
Gallaecio, line 15: Maybe we should mention this in the release notes as a bug fix. Before, `{{ foo }}` would have been silently removed, now it will raise an `UndefinedError` exception.

kmike (approved):

PR #118, opened 2024-11-12
feat: Migrated setup.py to pyproject.toml

### LINKED ISSUE #116: Migration from setup.py to pyproject.toml
Integrating the various tests, linters, formatters and docs generators to modern python using pyproject.toml

swayam0322:
Fixes #116 
- Migrated from setup.py to pyproject.toml 
- upgraded publish.yml file to use build instead of wheel 
- made changes to tox.ini to support pyproject.toml 
- removed setup.py and MANIFEST.in 
- added venv to the .gitignore 
- upgraded deprecated [bump2version](https://github.com/c4urself/bump2version) to [bump-my-version](https://github.com/callowayproject/bump-my-version) as per README


swayam0322:
@Gallaecio please review

.github/workflows/publish.yml
```diff
@@ -21,12 +21,12 @@ jobs:
...
+        pip install build twine
     - name: Build and publish
       if: startsWith(github.ref, 'refs/tags')
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
       run: |
-        python setup.py sdist bdist_wheel
+        python -m build
         twine upload dist/*
```
Gallaecio, line 32: Maybe we could simplify this code by copying it [from Scrapy](https://github.com/scrapy/scrapy/blob/master/.github/workflows/publish.yml).

pyproject.toml
```diff
@@ -8,13 +58,19 @@ multi_line_output = 3
...
 
 [tool.pytest.ini_options]
 filterwarnings = [
     'ignore:The zyte_common_items.ae module .*:DeprecationWarning',
-]
\ No newline at end of file
+]
+
+[tool.bumpversion]
+current_version = "0.27.0"
```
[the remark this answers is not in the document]
swayam0322: > You accidentally increased this version here.

It must have gotten updated accidentally while testing bump-my-version,

> Did you do it manually, or does this line get updated automatically without explicitly listing pyproject.toml in the list of files to edit below?

 the cli command  `bump-my-version (major/minor/patch)`  updates the version automatically in the .toml file and the VERSION file same as bump2version 

>  What happens if a dependency happens to have the same (minimum) version as the package, does it get edited as well? Do we need to implement some regular expression in the bump-my-version config to fix that?

 Its an updated fork of bump2version so I don't think it should work any different but would have to look further into it

swayam0322:
As per my understanding MANIFEST.in was used to include datafiles in the build 
but it was already added in the package data of the toml so I removed it 
```
[tool.setup_tools.package-data]
zyte_common_items = ["py.typed","VERSION"]
```

.github/workflows/publish.yml
swayam0322: Cool, will pick it up in the next commit

Gallaecio:
OK, I have tried running `python -m build` locally, and indeed `MANIFEST.in` does not seem necessary. Nice.

pyproject.toml
```diff
@@ -1,5 +1,55 @@
...
+changelog = "https://github.com/zytedata/zyte-common-items/blob/main/CHANGELOG.rst"
+
+[tool.setuptools.dynamic]
+version = { file = "zyte_common_items/VERSION" }
+
+[tool.setup_tools.package-data]
+zyte_common_items = ["py.typed","VERSION"]
+
+[tool.setuptools.packages.find]
+exclude= ["tests", "docs*"]
```
Gallaecio, line 49:
```suggestion
include= ["zyte_common_items"]
```

When trying `python -m build` locally, I got a file from the venv (`venv/bin/jp.py`). This change fixes that, and this is probably a safer approach to avoid including other unrelated files in the future.

swayam0322: Makes sense

Gallaecio (approved):

wRAR (approved):

Gallaecio:
Thanks!

PR #124, opened 2024-12-16
Add an item pipeline to log auto field stats

Gallaecio:
Reimplementation of https://github.com/scrapy-plugins/scrapy-zyte-api/pull/216 in zyte-common-items.

To do:
- [ ] Wait for @asadurski to test it out in a real project.

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -11,12 +11,50 @@ Installation
 
 
 .. _configuration:
+.. _scrapy-config:
 
-Configuration
-=============
+Scrapy configuration
+====================
 
-To allow itemadapter_ users, like Scrapy_, to interact with :ref:`items
-<items>`, prepend :class:`~zyte_common_items.ZyteItemAdapter` or
+-   If using Scrapy_ 2.10 or higher, enable the add-on:
+
+    .. code-block:: python
+        :caption: settings.py
+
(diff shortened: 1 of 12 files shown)


docs/setup.rst
```diff
@@ -11,12 +11,50 @@ Installation
...
+
+    -   Adds :class:`~zyte_common_items.pipelines.AutoFieldStatsItemPipeline`
+        (if :doc:`scrapy-poet <scrapy-poet:index>` is installed) and
+        :class:`~zyte_common_items.pipelines.DropLowProbabilityItemPipeline` to
+        :setting:`ITEM_PIPELINES <scrapy:ITEM_PIPELINES>`:
+
+        .. code-block:: python
+
+            ITEM_PIPELINES = {
+                "zyte_common_items.pipelines.DropLowProbabilityItemPipeline": 100,
```
kmike, line 43: I think we should improve the pipeline before enabling it by default in the addon.

In particular, it currently tries to process all items, and fails if an item is not an item defined in this package (or its subclass).

I also wonder if 0.1 default threshold is surprising or not. It was fine previously, when users needed to enable the pipeline explicitly. I'd probably expect the addon to configure the components, but not make such decisions - not sure.

wRAR (approved):

PR #125, opened 2024-12-20
Add support for custom attr results to DropLowProbabilityItemPipeline.

wRAR:


zyte_common_items/pipelines.py
```diff
@@ -115,25 +115,42 @@ def get_threshold_for_item(self, item, spider):
...
+        from scrapy.exceptions import DropItem
 
+        if isinstance(item, dict):
+            # for nested items remove sub-items that have low probability
+            # instead of dropping the whole result
+            new_item = {}
+            for item_type, sub_item in item.items():
+                if item_type == "customAttributes":
+                    new_item[item_type] = sub_item
+                    continue
```
kmike, line 141: Why do you need special handling for custom attributes? 
They don't have probability field, so they shouldn't be dropped?
Is it related to the fact the pipeline only supports zyte-common-items items, and breaks on arbitrary items? If so, it may be good to fix it as well, as it should (hopefully) simplify custom attributes implementation - they won't be a special case anymore (hopefully :)

wRAR: Two things: to simplify the flow somewhat (but I agree that with the current flow, which can process several sub-items, the flow is not simplified by this) and to not emit `drop_low_probability_item/{processed,kept}/customAttributes` stats which I thought are pointless (but harmless as there are just 2 of them). Should I remove the special handling?

kmike: I was probably thinking about a slightly different thing:

1. Pipeline should handle items with probability field - e.g. with zyte-common-items's item.get_probability(), and (maybe) by taking item["probability"] throiugh itemadapter.
2. For items without probability field, do nothing - don't drop them, don't change the stats.
3. There should be some special handling for nested items, where the exact items are inside a dictionary, probably with arbitrary keys.

I'm not sure how best to distinguish between (2) and (3) scenarios though; maybe it can be an explicit configuration, but maybe not.

With this approach customAttributes is just an item without probability field, so it doesn't need any special handling.

wRAR: We can omit stats for items where probability is None, or should we make some more changes?

kmike: Ideally, we should replace `item_proba = item.get_probability()` with something smarter, because it currently breaks on all non-zci items. It's possible to do it in another PR, but it seems pretty important to fix this, and it looked somewhat relevant here :)

wRAR: Should we just check if the item inherits `ProbabilityMixin` (or even `zyte_common_items.base.Item`?)?

kmike: This would work.
Sometimes users may be converting zci items to dicts before they reach the middleware, and it'd be nice to support probability in non-zci items in some way in general, but it looks like a larger change; just making sure we don't break completely is way better than what's currently in main.

Gallaecio (approved):

zyte_common_items/pipelines.py
```diff
@@ -115,25 +116,44 @@ def get_threshold_for_item(self, item, spider):
...
+            # for nested items remove sub-items that have low probability
+            # instead of dropping the whole result
+            new_item = {}
+            for item_type, sub_item in item.items():
+                threshold = self.get_threshold_for_item(sub_item, spider)
+                if self._process_probability(sub_item, threshold):
+                    new_item[item_type] = sub_item
+            if not new_item:
+                # everything has been removed
+                raise DropItem
```
kmike, line 150: Hm, there is a difference in error messages - should we copy some of the text from the DropItem below? 
We also don't log anything when sub-items are dropped; probably that's ok, but maybe not - what do you think?

Gallaecio: +1 to logging something when dropping sub-items.

zyte_common_items/pipelines.py
```diff
@@ -115,25 +119,54 @@ def get_threshold_for_item(self, item, spider):
...
+                if self._process_probability(sub_item, threshold):
+                    new_item[item_type] = sub_item
+                else:
+                    logger.info(
+                        f"This sub-item is dropped since the probability ({sub_item.get_probability()}) "
+                        f"is below the threshold ({threshold}):\n{sub_item!r}"
+                    )
+            if not new_item:
+                # everything has been removed
+                raise DropItem(
```
wRAR, line 160: Note that this needs updating after #126 merge

PR #126, opened 2024-12-30
Use a custom DropItem subclass

Gallaecio:
To enable custom logging to cleanly lower the message severity.

### DIFF
--- a/zyte_common_items/log_formatters.py
+++ b/zyte_common_items/log_formatters.py
@@ -1,0 +1,17 @@
+from logging import INFO
+
+from scrapy.exceptions import DropItem
+from scrapy.logformatter import LogFormatter
+
+
+class InfoDropItem(DropItem):
+    """DropItem subclass for items that should be dropped with an INFO message
+    (instead of the default WARNING message)."""
+
+
+class ZyteLogFormatter(LogFormatter):
+    def dropped(self, item, exception, response, spider):
+        data = super().dropped(item, exception, response, spider)
+        if isinstance(exception, InfoDropItem):
+            data["level"] = INFO
+        return data
--- a/zyte_common_items/pipelines.py
+++ b/zyte_common_items/pipelines.py
@@ -1,5 +1,10 @@
+# Raise ImportError if Scrapy is missing.
+import scrapy  # isort: skip  # noqa: F401
+
 from copy import deepcopy
 from warnings import warn
+
+from scrapy.exceptions import DropItem
 
 from zyte_common_items import ae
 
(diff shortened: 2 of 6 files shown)


zyte_common_items/pipelines.py
```diff
@@ -56,6 +61,11 @@ def process_item(self, item, spider):
         return ae.downgrade(item)
 
 
+class InfoDropItem(DropItem):
+    """DropItem subclass for items that should be dropped with an INFO message
+    (instead of the default WARNING message)."""
```
kmike, line 66: If this class is defined here, we should define the log formatter here as well (not in zyte-spider-templates-project).

Gallaecio: Am I correct to assume that we should also enable the new formatter in the library add-on, and not in the project?

kmike: Yes, it seems we should do it here.

Gallaecio: I meant the zyte-spider-templates library, given there is no add-on here. Or are you thinking we should add an add-on to zyte-common-items?

kmike: You're adding it in https://github.com/zytedata/zyte-common-items/pull/124 already :)

kmike (approved):

zyte_common_items/pipelines.py
kmike: As for the settings, it seems it'd be better to keep them in the project - as we discussed elsewhere, it might be simpler for addons not to configure other libraries. But I don't have a stong opinion on this.

zyte_common_items/log_formatters.py
```diff
@@ -0,0 +1,17 @@
...
+from scrapy.exceptions import DropItem
+from scrapy.logformatter import LogFormatter
+
+
+class InfoDropItem(DropItem):
+    """DropItem subclass for items that should be dropped with an INFO message
+    (instead of the default WARNING message)."""
+
+
+class ZyteLogFormatter(LogFormatter):
```
VMRuiz, line 12: Alternatively, we could allow to configure the severity level at Scrapy level via settings given that this is not the first time this has come up.

Gallaecio: @kmike mentioned somewhere that maybe we should allow components to set their desired log level in the exception, e.g. `raise DropItem(severity=INFO)`.

Maybe we could have a setting to implement the overall default log level, allow components to set a custom value through severity, and let components choose whether or not they respect the setting, force a given value, or provide their own setting to override their own request droppings.

Gallaecio: https://github.com/scrapy/scrapy/issues/6603

PR #127, opened 2024-12-31
Add an add-on, cover ZyteLogFormatter in the docs

Gallaecio:
Follow-up to https://github.com/zytedata/zyte-common-items/pull/126#discussion_r1899546111, with code from https://github.com/zytedata/zyte-common-items/pull/124.

### DIFF
--- a/docs/setup.rst
+++ b/docs/setup.rst
@@ -11,12 +11,46 @@ Installation
 
 
 .. _configuration:
+.. _scrapy-config:
 
-Configuration
-=============
+Scrapy configuration
+====================
 
-To allow itemadapter_ users, like Scrapy_, to interact with :ref:`items
-<items>`, prepend :class:`~zyte_common_items.ZyteItemAdapter` or
+If you use Scrapy, zyte-common-items provides some functionality that needs
+configuring:
+
+-   If using Scrapy_ 2.10 or higher, enable the add-on:
+
+    .. code-block:: python
+        :caption: settings.py
+
+        ADDONS = {
+            "zyte_common_items.Addon": 400,
+        }
+
+    The add-on:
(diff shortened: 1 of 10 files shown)


docs/setup.rst
```diff
@@ -11,12 +11,46 @@ Installation
...
+    needed.
+
+
+.. _itemadapter-config:
+
+itemadapter configuration
+=========================
+
+To allow itemadapter_ to interact with :ref:`items <items>`, prepend
+:class:`~zyte_common_items.ZyteItemAdapter` or
```
kmike, line 53: Should we mention addon here? It already does this configuration.

zyte_common_items/_addon.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Addon:
+    def update_settings(self, settings: BaseSettings) -> None:
+        if not any(
+            issubclass(cls, (ZyteItemAdapter, ZyteItemKeepEmptyAdapter))
+            for cls in ItemAdapter.ADAPTER_CLASSES
+        ):
+            ItemAdapter.ADAPTER_CLASSES = (ZyteItemAdapter,) + tuple(
+                ItemAdapter.ADAPTER_CLASSES
+            )
```
kmike, line 32: I think we might need to convert it to deque, because itemadapter docs [suggest](https://github.com/scrapy/itemadapter?tab=readme-ov-file#registering-an-adapter) using .appendleft: any code which tries to register an adapter class after the addon is executed would break, if it follows itemadapter docs.

That's too bad we're making the change globally - e.g. it's not possible to run 2 spiders in the same proces, with different adapter classes enabled. But that's definitely out of scope here :)

PR #128, opened 2025-01-14
0.27.0 release notes

Gallaecio:
@kmike Should we add the pipeline to the add-on, now that it supports nested items, or should we keep it off?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,20 @@
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


wRAR:
(good question)

kmike (approved):
That's fine to not enable DropLowProbability pipeline in the addon. I'm still worried a bit about the default of dropping items with probability < 0.1; it feels like something which should be done by the user explicitly.

PR #129, opened 2025-04-03
Migrate to ruff.

wRAR:


tests/__init__.py
```diff
@@ -1,37 +1,16 @@
-import contextlib
-import json
-import os
 import random
 
 
-def load_fixture(name):
```
wRAR, line 7: Both of these were unused.

PR #132, opened 2025-05-14
[CIT-1532] Add displayedUrlText to 

Nykakin:
A new `displayedUrlText` field to `SerpOrganicResult` was accepted as part of version 1.1 of Search Results schema.

### DIFF
--- a/zyte_common_items/items/serp.py
+++ b/zyte_common_items/items/serp.py
@@ -34,6 +34,9 @@ class SerpOrganicResult(Item):
     #: a search, must be 1.
     rank: Optional[int] = None
 
+    #: Result additional information such as number of instagram followers.
+    displayedUrlText: Optional[str] = None
+
 
 @attrs.define(kw_only=True)
 class SerpMetadata(ListMetadata):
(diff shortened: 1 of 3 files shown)


zyte_common_items/items/serp.py
```diff
@@ -34,6 +34,14 @@ class SerpOrganicResult(Item):
     #: a search, must be 1.
     rank: Optional[int] = None
 
+    #: Text displayed to represent :attr:`url`.
+    #:
+    #: It may not be an actual URL, but some stylized or simplified 
+    #: representation of it. For example, if :attr:`url` is
+    #: https://en.wikipedia.org/wiki/Foobar, :attr:`displayedUrlText` could be 
+    #: something like ``"https://en.wikipedia.org  › wiki › Foobar"``.
+    displayedUrlText: Optional[str] = None
```
Gallaecio:
```suggestion
    #: Text displayed to represent :attr:`url`.
    #:
    #: It may not be an actual URL, but some stylized or simplified
    #: representation of it. For example, if :attr:`url` is
    #: https://en.wikipedia.org/wiki/Foobar, :attr:`displayedUrlText` could be
    #: something like ``"https://en.wikipedia.org  › wiki › Foobar"``.
    displayedUrlText: Optional[str] = None
```

PR #141, opened 2025-09-26
Explicit reexport

wRAR:


zyte_common_items/pages/__init__.py
```diff
@@ -38,3 +38,53 @@
...
+    "BaseSearchRequestTemplatePage",
+    "BaseSerpPage",
+    "BaseSocialMediaPostPage",
+    "BusinessPlacePage",
+    "DescriptionMixin",
+    "ForumThreadPage",
+    "HasMetadata",
+    "JobPostingNavigationPage",
+    "JobPostingPage",
+    # "MetadataT",
```
wRAR, line 80: Not sure what to do with this (it was moved from `.pages.mixins` to `.components.metadata` in #92), we could drop it outright or deprecate this import first.

zyte_common_items/processors.py
```diff
@@ -124,7 +124,7 @@ def brand_processor(value: Any, page: Any) -> Any:
         value = value.strip()
         return Brand(name=value) if value else None
 
-    if isinstance(value, (Selector, SelectorList, HtmlElement)):
+    if isinstance(value, (Selector, HtmlElement)):
```
wRAR, line 127: bug, `value` can't be `SelectorList` after `value = _handle_selectorlist(value)`

Gallaecio (approved):

PR #142, opened 2025-10-03
Allow url=None in items.

### LINKED ISSUE #140: Don't raise an exception for .url attribute if it's set to None
We currently use `url_to_str` converter for the main item url attribute for various items (e.g. https://github.com/zytedata/zyte-common-items/blob/17efc3116f909aa7d6bc6709fc8b41ef8f5020e9/zyte_common_items/items/product.py#L570). 

While this is consistent with the typing, and with `url` attribute being required, it gets in a way with web-poet test fixture creation: it requires a working .url field implementation in a page object, not just a stub one which returns None.

While not perfect, as a middle-ground, I think we should change the converters to `url_to_str_optional` for all required .url attributes, and add instructions to ignore typing errors if they occur. The idea is to allow temporarily inconsistent state of the items (None url attribute), as it may be required for incremental item building, and as it makes it easier to work with web-poet test fixture creation workflow. 

The attribute itself should be kept required, and the declared type shouldn't change (i.e. no schema change - just remove the extra runtime validation).

wRAR:
Fixes #140

### DIFF
--- a/tests/test_items.py
+++ b/tests/test_items.py
@@ -992,5 +992,5 @@ def test_item_subclasses():
 )
 def test_url_none(cls: type) -> None:
     """Items can take url=None."""
-    item = cls(url=None)  # type: ignore[arg-type]
+    item = cls(url=None)
     assert item.url is None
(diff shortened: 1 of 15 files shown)


Gallaecio (approved):
:rocket:

kmike (approved):

